PR #21, opened 2022-03-18
Update references to Private Aggregation API

alexmturner:
Also updates examples to the syntax proposed in the API's explainer


alexmturner:
@pythagoraskitty could you PTAL? I changed up the example slightly to better match the explainer, but we could maybe keep the campaign ID around -- the `sendCountDistinctReport()` call is little bit of an extension anyway, so we could maybe extend it once more.

PR #23, opened 2022-04-12
Update reach example

pythagoraskitty:


README.md
```diff
@@ -162,14 +162,17 @@ Worklet script (i.e. `reach.js`):
 ```js
 class SendReachReportOperation {
   async function run(data) {
-    // A toy model that only computes reach for users whose favorite color is red.
-    if (data["favorite-color"] != "red") {
+    // A toy model that only computes reach for users who haven't previously had a report sent.
+    if (await this.sharedStorage.get("report-sent") != "yes") {
```
alexmturner, line 166: I'm wondering if this "report sent" value should be scoped per campaign to measure per-campaign reach. (We could also consider fixing the bucket below instead so that this measures overall reach for all campaigns.)

README.md
```diff
@@ -162,14 +162,17 @@ Worklet script (i.e. `reach.js`):
 ```js
 class SendReachReportOperation {
   async function run(data) {
-    // A toy model that only computes reach for users whose favorite color is red.
-    if (data["favorite-color"] != "red") {
+    // A toy model that only computes reach for users who haven't previously had a report sent.
```
alexmturner, line 165: I'm not sure that the toy model description still applies here. Maybe just something along the lines of "To calculate reach, we skip users who have already had a report sent."?

Alternatively, if we want to keep the toy model description, we could have the if statement be something like `if (await this.sharedStorage.get("report-sent") != "yes") && data["favorite-color"] != "red") {`

Wdyt?

README.md
```diff
@@ -162,14 +162,17 @@ Worklet script (i.e. `reach.js`):
...
+    if (await this.sharedStorage.get("report-sent") != "yes") {
       return;  // Don't send a report.
     }
 
     // The user agent will send the report to a default endpoint after a delay.
-    privateAggregation.sendCountDistinctReport({
-      bucket: (await this.sharedStorage.get("id"))});
+    privateAggregation.sendHistogramReport({
+      bucket: (await this.sharedStorage.get("id")),
+      key: data["campaign-id"]});
```
alexmturner, line 173: `sendHistogramReport()` takes a bucket and value. I think we'd probably want the campaign ID as the bucket if we're measuring per-campaign reach (or some fixed bucket if we're measuring overall reach). The value would then be a constant (see [here](https://github.com/alexmturner/private-aggregation-api#:~:text=convertAgeToBucketId(demo%5B%22age%22%5D)%3B-,value%3A%20128,-%2C%20%20//%20A%20predetermined%20fixed) for an example). I don't think we need the per-browser ID any more as we're directly preventing duplicate reports.

README.md
pythagoraskitty: Never mind, I just reread what you wrote, and the second part makes more sense now.  :)

README.md
pythagoraskitty: I can take out "toy model", no problem.  Will do!

pythagoraskitty:
OK, I think that I incorporated what you said.  Would you please take a last look?  Thanks.

README.md
```diff
@@ -162,14 +160,22 @@ Worklet script (i.e. `reach.js`):
...
       return;  // Don't send a report.
     }
 
     // The user agent will send the report to a default endpoint after a delay.
-    privateAggregation.sendCountDistinctReport({
-      bucket: (await this.sharedStorage.get("id"))});
+    privateAggregation.sendHistogramReport({
+      bucket: data["campaign-id"];
+      value: 128,  // A predetermined fixed value; see Private Aggregation API explainer: Scaling values.
+      });
```
alexmturner, line 176: nit: unindent two spaces

README.md
```diff
@@ -162,14 +160,22 @@ Worklet script (i.e. `reach.js`):
 ```js
 class SendReachReportOperation {
   async function run(data) {
-    // A toy model that only computes reach for users whose favorite color is red.
-    if (data["favorite-color"] != "red") {
+    const report_sent_for_campaign = "report-sent-" + data["campaign-id"];
+    
+    // Compute reach only for users who haven't previously had a report sent for this campaign.
```
alexmturner, line 165: minor: Arguably we're still computing reach for the skipped users -- it's just already complete. What do you think about something like "To compute reach, we skip users who have already been counted. That is, those who have previously sent a report for this campaign will be skipped, even if it was triggered by a site other than the current one."?

alexmturner (approved):
LGTM, thanks! Just two optional thoughts

PR #28, opened 2022-05-18
use dictionaries instead of URLs in selectURL()

pythagoraskitty:
[xyaoinum](https://github.com/xyaoinum) Could you please take a look?  Thank you!


pythagoraskitty:
jkarlin@, Could you please take a look?  Thank you!

PR #38, opened 2022-07-21
Update README.md

etrouton:
Additional information about our data retention policy and our data storage limits

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,12 @@ After a document dies, the corresponding worklet will be kept alive for maximum 
 
 Shared storage methods can be disallowed by the "shared-storage" [policy-controlled feature](https://w3c.github.io/webappsec-permissions-policy/#policy-controlled-feature). Its default allowlist is * (i.e. every origin). 
 
+## Data Retention Policy
+Users’ shared storage data is cleared every 30 days from time of creation on a per-origin basis.
+
+## Data Storage Limits
+Shared Storage allows each origin up to 10,000 key/value pairs, with each key and value limited to a maximum of 1024 char16_t characters apiece. 
+
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -244,6 +244,12 @@ After a document dies, the corresponding worklet will be kept alive for maximum
 
 Shared storage methods can be disallowed by the "shared-storage" [policy-controlled feature](https://w3c.github.io/webappsec-permissions-policy/#policy-controlled-feature). Its default allowlist is * (i.e. every origin). 
 
+## Data Retention Policy
+Users’ shared storage data is cleared every 30 days from time of creation on a per-origin basis.
+
+## Data Storage Limits
+Shared Storage allows each origin up to 10,000 key/value pairs, with each key and value limited to a maximum of 1024 char16_t characters apiece.
```
jkarlin, line 251: Suggest removing 'char16_t' here.

etrouton: Good idea, thanks Josh.

jkarlin: Suggest adding a bit of rationale for the limit. Something along the lines of, "Shared Storage is not subject to the quota manager, as that would leak information across sites. Therefore we limit its size in the following way:" or some such.

README.md
```diff
@@ -244,6 +244,12 @@ After a document dies, the corresponding worklet will be kept alive for maximum
 
 Shared storage methods can be disallowed by the "shared-storage" [policy-controlled feature](https://w3c.github.io/webappsec-permissions-policy/#policy-controlled-feature). Its default allowlist is * (i.e. every origin). 
 
+## Data Retention Policy
+Users’ shared storage data is cleared every 30 days from time of creation on a per-origin basis.
```
jkarlin, line 248: Suggest adding justification for this.

PR #40, opened 2022-09-07
Rough spec

pythagoraskitty:
I don't actually want to land this yet, because it's a very rough WIP.  But posting it here to help with collaboration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,22 @@
+name: Build
+on:
+  pull_request:
+    branches:
+    - main
+  push:
+    branches:
+    - main
+jobs:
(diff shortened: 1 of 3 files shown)


pythagoraskitty:
This rough spec skeleton is still incomplete, but wanderview@ has requested that I land something on the main branch so that the workflow starts taking effect and it makes it easier to review.  xyaoinum@, PTAL. Thanks!

pythagoraskitty:
This rough spec skeleton is still incomplete, but wanderview@ has requested that I land something on the main branch so that the workflow starts taking effect and it makes it easier to review.  xyaoinum@, PTAL. Thanks!

PR #41, opened 2022-09-07
Update README.md

pythagoraskitty:
What exactly needs to be done here, given that this is an API change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ A third-party, `a.example`, wants to randomly assign users to different groups (
 
 To do so, `a.example` writes a seed to its shared storage (which is not added if already present). `a.example` then registers and runs an operation in the shared storage [worklet](https://developer.mozilla.org/en-US/docs/Web/API/Worklet) that assigns the user to a group based on the seed and the experiment name and chooses the appropriate ad for that group.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,6 +135,8 @@ There have been multiple privacy proposals ([SPURFOWL](https://github.com/AdRoll
     *   Returns a promise that resolves into the `n`th key or the number of keys, respectively.
 *   `sharedStorage.set(key, value, options)`, `sharedStorage.append(key, value)`, `sharedStorage.delete(key)`, and `sharedStorage.clear()`
     *   Same as outside the worklet, except that the promise returned only resolves into `undefined` when the operation has completed.
+*   `sharedStorage.remainingBudget()`
+    *   Returns a double indicating the remaining available privacy budget for `sharedStorage.selectURL()`, in bits.
```
jkarlin, line 139: JavaScript only has one number type (64-bit float), called number.

pythagoraskitty: Changed.

pythagoraskitty:
Per the new guidance on what counts as a significant change, I have marked this as substantive, since it's really only giving developers more detail about the budgeting process that is previously described.

PR #42, opened 2022-09-07
Update README.md

pythagoraskitty:
I noticed that we never updated the API to reflect the change to using async iterators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ There have been multiple privacy proposals ([SPURFOWL](https://github.com/AdRoll
 *   `sharedStorage.length()`
     *   Returns a promise that resolves into the number of keys.
 *   `sharedStorage.keys()` and `sharedStorage.entries()`
-    *   Returns an object that can iterate through the origin's keys or key-value pairs, respectively, using the `next()` method.
+    *   Returns an object that can iterate asynchronously through the origin's keys or key-value pairs, respectively, using the `next()` method.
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
xyaonium@, PTAL. Thanks!

README.md
```diff
@@ -131,8 +131,10 @@ There have been multiple privacy proposals ([SPURFOWL](https://github.com/AdRoll
 
 *   `sharedStorage.get(key)`
     *   Returns a promise that resolves into the `key`‘s entry or an empty string if the `key` is not present.
-*   `sharedStorage.key(n)` and `sharedStorage.length()`
-    *   Returns a promise that resolves into the `n`th key or the number of keys, respectively.
+*   `sharedStorage.length()`
+    *   Returns a promise that resolves into the number of keys.
+*   `sharedStorage.keys()` and `sharedStorage.entries()`
+    *   Returns an object that can iterate asynchronously through the origin's keys or key-value pairs, respectively, using the `next()` method.
```
xyaoinum, line 137: (Looked over the [KV Storage explainer](e.g. https://github.com/WICG/kv-storage/tree/7c553add86e6166820c29acf79abfc23ebbf8737#keys))
The async iterator seems to be a publicly known concept, so I think we can just follow the pattern there, e.g. "Returns an [async iterator](https://webidl.spec.whatwg.org/#es-asynchronous-iterator-prototype-object) for all the stored keys or [key, value] pairs, sorted in the underlying key order."

pythagoraskitty: Done.  Is it ok now?

PR #51, opened 2022-12-13
Make sharedStorage aware of size + hook into fenced frame changes

gtanzer:
See https://github.com/WICG/fenced-frame/pull/56, and analogous changes to FLEDGE in https://github.com/WICG/turtledove/pull/417 .

Basically, now you need to declare a size across all urls to be selected, and that size is part of the k-anonymity check that was previously just the URL.

This has a number of benefits, like:
* **Privacy:** In the old design, there was wiggle room for the embedder to cause observable differences inside the fenced frame by its choice of the size. In the new design, we close this gap by running the k-anonymity check on the totality of information that flows from the embedder into the fenced frame.
* **Flexibility:** Due to the privacy improvements above, we no longer need to stipulate a list of allowed sizes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ const opaqueURL = await window.sharedStorage.selectURL(
     {url: "blob:https://b.example/abc…", reportingMetadata: {"click": "https://report.example/a..."}},
     {url: "blob:https://c.example/789…"}
   ],
-  { data: { name: 'experimentA' } }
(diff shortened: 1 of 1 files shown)


gtanzer:
In light of the changes in https://github.com/WICG/shared-storage/pull/69, these size changes are no longer necessary. Instead, all size restrictions (i.e. size freezing, a size allowist) for selectURL fenced frames will be removed, as any information that enters the fenced frame through the size could have instead been communicated in through the url.

PR #55, opened 2023-02-02
Update spec.bs draft 

pythagoraskitty:
Updating spec.bs to the latest rough draft.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -6,57 +6,417 @@ Status: CG-DRAFT
 Group: WICG
 URL: https://github.com/WICG/shared-storage
 Editor: Camillia Smith Barnes, Google https://google.com, cammie@chromium.org
+Markup Shorthands: markdown yes
 Abstract: Shared Storage is a storage API that is intentionally not partitioned by top-frame site (though still partitioned by context origin of course!). To limit cross-site reidentification of users, data in Shared Storage may only be read in a restricted environment that has carefully constructed output gates.
 </pre>
+<pre class=link-defaults>
+spec:url; type:interface; text:URL
+</pre>
+<style>
+/* .XXX from https://resources.whatwg.org/standard.css */
+.XXX {
+  color: #D50606;
+  background: white;
+  border: solid #D50606;
+}
+</style>
 
 Introduction {#intro}
 =====================
 <em>This section is not normative.</em>
 
-This document introduces a new primitive.
+In order to prevent cross-site user tracking, browsers are partitioning all forms of storage by top-frame site. But, there are many legitimate use cases currently relying on unpartitioned storage.
 
-<span class=XXX>TODO: Finish adding interfaces.</span>
-<span class=XXX>TODO: Add algorithms for APIs.</span>
-<span class=XXX>TODO: Add text, including fleshing out intro.</span>
+This document introduces a new storage API that is intentionally not partitioned by top-frame site (though still partitioned by context origin), in order to serve a number of the use cases needing unpartitioned storage. To limit cross-site reidentification of users, data in Shared Storage may only be read in a restricted environment that has carefully constructed output gates. Over time, we hope to design and add additional gates.
 
-Shared Storage DOM Interface {#dom-interface}
-=============================================
+<div class="example">
+  `a.example` randomly assigns users to groups in a way that is consistent cross-site.
+
+  Inside an `a.example` iframe:
+  <pre class="lang-js">
+    function generateSeed() { … }
+    await window.sharedStorage.worklet.addModule('experiment.js');
+
+    // Only write a cross-site seed to a.example's storage if there isn't one yet.
+    window.sharedStorage.set('seed', generateSeed(), { ignoreIfPresent: true });
+
+    // opaqueURL will be of the form urn:uuid and will be created by privileged code to
+    // avoid leaking the chosen input URL back to the document.
+
+    const opaqueURL = await window.sharedStorage.selectURL(
+      'select-url-for-experiment',
+      [
+        {url: "blob:https://a.example/123…", reportingMetadata: {"click": "https://report.example/1..."}},
+        {url: "blob:https://b.example/abc…", reportingMetadata: {"click": "https://report.example/a..."}},
+        {url: "blob:https://c.example/789…"}
+      ],
+      { data: { name: 'experimentA' } }
+    );
+
+    document.getElementById('my-fenced-frame').src = opaqueURL;
+  </pre>
+
+  inside the `experiment.js` worklet script:
+  <pre class="lang-js">
+    class SelectURLOperation {
+      hash(experimentName, seed) { … }
+
+      async run(data, urls) {
+        const seed = await this.sharedStorage.get('seed');
+        return hash(data.name, seed) % urls.length;
+      }
+    }
+    register('select-url-for-experiment', SelectURLOperation);
+  </pre>
+</div>
+
+The {{SharedStorageWorklet}} Interface {#worklet}
+=================================================
 <xmp class='idl'>
+  [Exposed=(Window)]
+  interface SharedStorageWorklet : Worklet {
+  };
+</xmp>
 
-[Exposed=(Window)]
-interface SharedStorage {
-  Promise<any> set(DOMString key, DOMString value, optional SharedStorageSetMethodOptions options = {});
+The {{SharedStorageWorklet}} object allows developers to supply scripts to process Shared Storage data on a separate thread and then output the result through one or more of the output gates. Currently there are two output gates, the private aggregation output gate and the URL selection output gate.
 
-  Promise<any> append(DOMString key, DOMString value);
+The {{SharedStorageWorklet}}'s worklet global scope type is {{SharedStorageWorkletGlobalScope}}. The scripts should define and {{register}} one or more {{SharedStorageOperation}}s. 
 
-  Promise<any> delete(DOMString key);
+<xmp class='idl'>
+  [Exposed=SharedStorageWorklet, Global=SharedStorageWorklet]
+  interface SharedStorageWorkletGlobalScope : WorkletGlobalScope {
+    undefined register(DOMString name,
+                       SharedStorageOperationConstructor operationCtor);
+  };
 
-  Promise<any> clear();
+  callback SharedStorageOperationConstructor = SharedStorageOperation(object options);
 
-  Promise<USVString> selectURL(DOMString name,
+  [Exposed=SharedStorageWorklet]
+  interface SharedStorageOperation {
+  };
+</xmp>
+
+Currently each {{SharedStorageOperation}} registered via {{SharedStorageWorkletGlobalScope/register()}} must be one of the following two types:
+* {{SharedStorageRunOperation}}
+* {{SharedStorageSelectURLOperation}}
+
+A {{SharedStorageRunOperation}} is an {{SharedStorageOperation}} that has no JavaScript return value, and hence is especially suited for calling into the private aggregation service.
+
+A {{SharedStorageSelectURLOperation}} is an {{SharedStorageOperation}} that takes in a list of {{SharedStorageUrlWithMetadata}} dictionaries (i.e. URLs each wrapped with metadata) and returns a {{long}} integer index specifying which of these URLs should be selected.
+
+<xmp class='idl'>
+  [Exposed=SharedStorageWorklet]
+  interface SharedStorageRunOperation : SharedStorageOperation {
+    Promise<undefined> run(object data);
+  };
+
+  [Exposed=SharedStorageWorklet]
+  interface SharedStorageSelectURLOperation : SharedStorageOperation {
+    Promise<long> run(object data,
+                      FrozenArray<SharedStorageUrlWithMetadata> urls);
+  };
+
+  dictionary SharedStorageUrlWithMetadata {
+    USVString url;
+    object reportingMetadata;
+  };
+</xmp>
+
+  ## Registering a Worklet ## {#register-worklet}
+
+  <div algorithm>
+    The <dfn method for="SharedStorageWorkletGlobalScope">register(|name|, |operationCtor|)</dfn> method steps are:
+
+    1. If |name| is missing or empty, throw a {{TypeError}}.
+    1. Let |operation_map| be the internal operation <a href="https://infra.spec.whatwg.org/#ordered-map">map</a> for this scope.
+    1. If |name| already exists as a key in |operation_map|, throw a {{TypeError}}.
+    1. If |operationCtor| is missing or not a constructor, throw a {{TypeError}}.
+    1. Let |prototype| be the class prototype for |operationCtor|.
+    1. If |prototype| is not an object, throw a {{TypeError}}.
+    1. Let |run| be the "run" method in |prototype|.
+    1. If |run| is missing or not a function, throw a {{TypeError}}.
+    1. <a href="https://infra.spec.whatwg.org/#map-set">Set</a> the value of |name| to |run| in |operation_map|.
+  </div>
+
+The {{SharedStorage}} Interface {#shared-storage}
+=================================================
+The {{SharedStorage}} interface is exposed to both the {{Window}} and the {{SharedStorageWorklet}}, with some of its methods only exposed to one or the other.
+
+Methods that allow the setting and/or deleting of data are exposed to both the {{Window}} and the {{SharedStorageWorklet}}. This makes it possible to modify the data in Shared Storage from multiple contexts.
+
+Meanwhile, methods for running {{SharedStorageOperation}}s, along with the {{SharedStorage/worklet}} attribute which is used to call {{Worklet/addModule()}}, are exposed to the {{Window}} only, as these are the means by which the {{Window}} interacts with the {{SharedStorageWorklet}}.
+
+In order to carefully control the flow of data read from Shared Storage, methods for getting data from the database backend are exposed to the {{SharedStorageWorklet}} only.
+
+Because bits of entropy can leak via {{SharedStorage/selectURL()}} if a user activates the fenced frame that navigates to the resulting opaque URL, we impose on each origin a <dfn>bit budget</dfn>, i.e. a maximum allowance of entropy bits that are permitted to leak during a given [=epoch=], where we define an <dfn>epoch</dfn> as a time period whose duration is a fixed pretermined length. When a leak occurs, its value in entropy bits is calculated and stored as a <dfn>bit debit</dfn> with a timestamp, and only [=bit debit=]s whose timestamps are within the current [=epoch=] are used to determine an origin's remaining budget. When an origin has insufficient remaining budget, {{SharedStorage/selectURL()}} will return its default value.
+
+<xmp class='idl'>
+  [Exposed=(Window,SharedStorageWorklet)]
+  interface SharedStorage {
+    [Exposed=(Window)] 
+    Promise<any> run(DOMString name,
+                     optional SharedStorageRunOperationMethodOptions options = {});
+
+    [Exposed=(Window)] 
+    Promise<USVString> selectURL(DOMString name,
                                  FrozenArray<SharedStorageUrlWithMetadata> urls,
                                  optional SharedStorageRunOperationMethodOptions options = {});
 
-  Promise<any> run(DOMString name, optional SharedStorageRunOperationMethodOptions options = {});
+    Promise<any> set(DOMString key,
+                     DOMString value,
+                     optional SharedStorageSetMethodOptions options = {});
 
-  readonly attribute SharedStorageWorklet worklet;
-};
+    Promise<any> append(DOMString key,
+                        DOMString value);
 
-[Exposed=(Window)]
-interface SharedStorageWorklet {
-  Promise<undefined> addModule(USVString moduleURL);
-};
+    Promise<any> delete(DOMString key);
 
-dictionary SharedStorageSetMethodOptions {
-  boolean ignoreIfPresent;
-};
+    Promise<any> clear();
 
-dictionary SharedStorageRunOperationMethodOptions {
-  object data;
-};
+    [Exposed=(SharedStorageWorklet)] Promise<DOMString> get(DOMString key);
 
-dictionary SharedStorageUrlWithMetadata {
-  required USVString url;
-  object reportingMetadata;
-};
+    [Exposed=(SharedStorageWorklet)] Promise<unsigned long> length();
+
+    [Exposed=(SharedStorageWorklet)] Promise<double> remainingBudget();
+
+    [Exposed=(SharedStorageWorklet)] SharedStorageIterator keys();
+
+    [Exposed=(SharedStorageWorklet)] SharedStorageIterator entries();
+
+    [Exposed=(Window)] readonly attribute SharedStorageWorklet worklet;
+  };
+
+  dictionary SharedStorageSetMethodOptions {
+    boolean ignoreIfPresent;
+  };
+
+  dictionary SharedStorageRunOperationMethodOptions {
+    object data;
+  };
+
+  enum IteratorMode { "key", "keyValue" };
+
+  [Exposed=(SharedStorageWorklet)]
+  interface SharedStorageIterator {
+    constructor(IteratorMode mode);
+    Promise<object> next();
+  };
 </xmp>
+
+  ## Note on Parallelism ## {#parallelism}
+
+  The {{SharedStorage}} interface runs its methods <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel">in parallel</a>. To avoid race conditions within a single context, each context that calls into the interface will have its own <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> for handling operations.
+
+  In particular, the {{Window}}'s {{SharedStorage}} operations for a given frame will use a separate <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> from that used by any {{SharedStorageWorklet}} that may be spun up by this frame.
+
+  When an algorithm before asks for steps to be enqueued, these steps should be enqueued on the <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> of the calling context, unless otherwise specified.
+
+  ## Run Operation Methods ## {#run-op}
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">run(|name|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If {{Worklet/addModule()}} has not yet been called, reject |promise|.
+        1. Let |worklet_queue| be the <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> for the context's associated {{SharedStorageWorklet}} interface, if it exists.
+        1. If |worklet_queue| does not exist, create it.
+        1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a> on |worklet_queue|:
+            1. If module script has been loaded for this worklet:
+                1. Let |operation_map| be the operation map in the associated {{SharedStorageWorkletGlobalScope}}.
+                1. Let |operation| be the operation registered to |name| in |operation_map|, if |operation| exists.
+                1. If |operation| exists:
+                    1. If |options| contains |data|, invoke |operation| with argument |data|.
+                    1. Otherwise, invoke |operation| with an empty argument list.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">selectURL(|name|, |urls|, |options|)</dfn> method steps are:
+
+    1. Let |config_promise| be a new promise.
+    1. If |urls| is empty or exceeds the maximum allowed length, reject |config_promise|.
+    1. Let |url_list| be an empty {{list}}.
+    1. For each |url_with_metadata| in |urls|:
+        1. If |url_with_metadata| has no field "|url|", reject |config_promise|.
+        1. Otherwise, let |url_string| be |url_with_metadata|["|url|"].
+        1. Let |url| be the result of running a <a href="https://url.spec.whatwg.org/#concept-url-parser">URL parser</a> on |url_string|.
+        1. If |url| is not a valid {{URL}}, reject |config_promise|.
+        1. Otherwise, [=list/append=] |url_string| to |url_list|.
+        1. If |url_with_metadata| has field "|reportingMetadata|":
+            1. Let |reporting_metadata| be |url_with_metadata|["|reportingMetadata|"].
+            1. If |reporting_metadata| is invalid, reject |config_promise|.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If {{Worklet/addModule()}} has not yet been called, reject |config_promise|.
+        1. Let |pending_url| be a newly generated <a href="https://www.rfc-editor.org/rfc/rfc4122">urn:uuid</a>.
+        1. Let |fenced_frame_config| be a <a href="https://wicg.github.io/fenced-frame/fencedframeconfig">FencedFrameConfig</a> constructed from |pending_url|.
+        1. Resolve |config_promise| as |fenced_frame_config|.
+        1. Let |index_promise| be a new promise.
+        1. Let |worklet_queue| be the <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> for the context's associated {{SharedStorageWorklet}} interface, if it exists.
+        1. If |worklet_queue| does not exist, create it.
+        1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a> on |worklet_queue|            1. Let |index| be 0.
+            1. If module script has been loaded for this worklet:
+                1. Let |operation_map| be the operation map in the associated {{SharedStorageWorkletGlobalScope}}.
+                1. Let |operation| be the operation registered to |name| in |operation_map|, if |operation| exists.
+                1. If |operation| exists:
+                    1. If |options| contains |data|, let |operation_result| be the result of invoking |operation| with arguments |url_list| and |data|.
+                    1. Otherwise, let |operation_result| be the result of invoking |operation| with argument |url_list|.
+                    1. If there are no errors:
+                        1. Set |index| to the result of casting |operation_result| to an {{unsigned long}}. If this throws an exception, catch it and set |index| back to 0.
+                        1. If |index| > |url_list|.[=list/size=], then set |index| back to 0.
+            1. Resolve |index_promise| with |index|.
+        1. When |index_promise| is resolved, let |result_index| be its numerical value.
+        1. Let |remaining_budget| be the numerical result from {{SharedStorage/remainingBudget()}}.
+        1. Let |current_debit| be log_2(|url_list|.[=list/size=]).
+        1. If |current_debit| > |remaining_budget|, set |result_index| to 0.
+        1. Store |current_debit| so that it can be withdrawn from the database in the event of a frame created from |fenced_frame_config| triggering a top-frame navigation.
+        1. Internally map |pending_url| to |url_list|[|result_index|].
+  </div>
+
+  ## Setter/Deleter Methods ## {#setter}
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">set(|key|, |value|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |key| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |value| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is empty, create |options| with "|ignoreIfPresent|" set to false.
+        1. Perform a database query to determine whether an |entry| for |key| already exists in the database for this context's origin.
+        1. If |entry| exists:
+            1. If |options|["|ignoreIfPresent|"] is true, resolve |promise| as undefined.
+            1. Otherwise, perform a database query to update |entry| with |key| and |value|.
+            1. Resolve |promise| as undefined.
+        1. Otherwise, perform a database query to insert an |entry| with |key| and |value| into the database for this context's origin.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">append(|key|, |value|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |key| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |value| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. Perform a database query to determine whether an |entry| for |key| already exists in the database for this context's origin.
+        1. If |entry| exists:
+            1. Let |current_value| be the current value of |entry|.
+            1. Perform a database query to update |entry| with |key| and (|current_value| + |value|).
+            1. Resolve |promise| as undefined.
+        1. Otherwise, perform a database query to insert an |entry| with |key| and |value| into the database for this context's origin.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">delete(|key|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |key| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. Perform a database query to delete any entry for |key| in the database for this context's origin.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">clear()</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. Perform a database query to delete all entries in the database for this context's origin.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  ## Getter Methods ## {#getter}
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">get(|key|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |key| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. Perform a database query to determine whether an |entry| for |key| exists in the database for this context's origin.
+        1. If |entry| exists:
+            1. Let |value| be the value of |entry|.
+            1. Resolve |promise| as |value|.
+        1. Otherwise, resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">length()</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. Let |num_entries| be the number of entries in the database for this context's origin, as determined by a database query to count them.
+        1. Resolve |promise| as |num_entries|.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">remainingBudget()</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. Let |total_budget| be the allotted total [=bit budget=] per [=epoch=].
+        1. Perform a database query to retrieve the |debit_sum| for the context's origin, i.e. the sum of the [=bit debit=]s whose timestamps are within the current [=epoch=].
+        1. Resolve |promise| as (|total_budget| - |debit_sum|).
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">keys()</dfn> method steps are:
+
+    1. Let |mode| be "{{IteratorMode/key}}".
+    1. Construct a {{SharedStorageIterator}} object |iterator| with |mode|.
+    1. Return |iterator|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">entries()</dfn> method steps are:
+
+    1. Let |mode| be "{{IteratorMode/keyValue}}".
+    1. Construct a {{SharedStorageIterator}} object |iterator| with |mode|.
+    1. Return |iterator|.
+  </div>
+
+  ## {{SharedStorageIterator}} Interface Methods ## {#iterator}
+
+  <div algorithm>
+    The <dfn method for="SharedStorageIterator">constructor(|mode|)</dfn> method steps are:
+
+    1. Create an internal queue |pending_entries|.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |mode| is "{{IteratorMode/key}}":
+            1. Query the database for all |keys| with context origin.
+            1. Queue |keys| in |pending_entries|, to be returned individually with the {{SharedStorageIterator/next()}} method.
+        1. Otherwise:
+            1. Query the database for all key-value pairs |entries| with context origin.
+            1. Queue |entries| in |pending_entries|, to be returned individually with the {{SharedStorageIterator/next()}} method.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="SharedStorageIterator">next()</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If there was an error in retrieving entries, reject |promise|.
+        1. If the internal entry queue |pending_entries| is empty:
+            1. Create an object |done_object| flagging that iteration has finished.
+            1. Resolve |promise| with |done_object|.
+        1. Otherwise, let |entry| be the entry at the front of |pending_entries|.
+        1. Resolve |promise| as |entry|.
+    1. Return |promise|.
+  </div>
+
+Privacy Considerations {#privacy}
+=================================================
+<p>
+  <span class=XXX>TODO: Add privacy considerations.</span>
+</p>
--- a/spec.bs
+++ b/spec.bs
@@ -119,6 +119,7 @@ spec: html; urlPrefix: https://www.rfc-editor.org/rfc/rfc4122
         text: urn:uuid
 spec: html; urlPrefix: https://wicg.github.io/fenced-frame/
     type: dfn
+        text: fenced frame; url: the-fencedframe-element
         text: FencedFrameConfig; url: fencedframeconfig
         text: url; for: FencedFrameConfig; url: dom-fencedframeconfig-url
 spec: md; urlPrefix: https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/README.md
@@ -158,11 +159,9 @@ Introduction {#intro}
 =====================
 <em>This section is not normative.</em>
 
-In order to prevent cross-site user tracking, browsers are partitioning all forms of storage by top-frame site; see .[=Client-Side Storage Partitioning=]. But, there are many legitimate use cases currently relying on unpartitioned storage.
+In order to prevent cross-site user tracking, browsers are partitioning all forms of storage by top-frame site; see [=Client-Side Storage Partitioning=]. But, there are many legitimate use cases currently relying on unpartitioned storage.
 
-This document introduces a new storage API that is intentionally not partitioned by top-frame site (though still partitioned by context origin), in order to serve a number of the use cases needing unpartitioned storage. To limit cross-site reidentification of users, data in Shared Storage may only be read in a restricted environment that has carefully constructed output gates. Over time, there may be additional ouput gates included in the standard.
-
-<span class=todo>Talk about Fenced Frames.</span>
+This document introduces a new storage API that is intentionally not partitioned by top-frame site (though still partitioned by context origin), in order to serve a number of the use cases needing unpartitioned storage. To limit cross-site reidentification of users, data in Shared Storage may only be read in a restricted environment, such as in a [=fenced frame=] or via the [=private aggregation|private aggregation service=]. Over time, there may be additional ouput gates included in the standard.
 
 <div class="example">
   `a.example` randomly assigns users to groups in a way that is consistent cross-site.
@@ -185,6 +184,7 @@ This document introduces a new storage API that is intentionally not partitioned
       { data: { name: 'experimentA' } }
     );
 
+    // Assumes that the fenced frame 'my-fenced-frame' has already been attached.
     document.getElementById('my-fenced-frame').config = fencedFrameConfig;
   </pre>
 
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
jkarlin@, could you please take a look when you get a chance?  Thanks.

I will also ask wanderview@ to take a look once he accepts the invite to the repo.

pythagoraskitty:
xyaoinum@, would you please take a look? Thanks.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+<style>
+/* .XXX from https://resources.whatwg.org/standard.css */
+.XXX {
+  color: #D50606;
+  background: white;
+  border: solid #D50606;
+}
+</style>
 
 Introduction {#intro}
```
xyaoinum, line 24: Some general style suggestions:

- Use |camelCase| for temporary variables
- Defined types and fields should be <dfn>spaced out naturally</dfn>
- When you define something that is associated with another thing, use <dfn for="shared storage data">per origin data</dfn>, and access it via: Let |perOriginData| be |sharedStorageData|’s [=shared storage data/per origin data=].

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+  enum IteratorMode { "key", "keyValue" };
+
+  [Exposed=(SharedStorageWorklet)]
+  interface SharedStorageIterator {
+    constructor(IteratorMode mode);
+    Promise<object> next();
+  };
 </xmp>
+
+  ## Note on Parallelism ## {#parallelism}
```
xyaoinum, line 206: The worklet spec already has the notion of **agent** for a worklet: e.g. https://html.spec.whatwg.org/multipage/worklets.html#create-a-worklet-global-scope. It’s roughly a "thread". Perhaps we still need to define more about the process/thread model (i.e. the implementation defined part), but at least it’s better to relate to the existing terms, so that we don’t need to define a separate task queue for the worklet operations. e.g. for run() and selectURL(), after the initial sanity checks, we can:
1. Let outsideSettings be worklet's relevant settings object.
2. Let agent be the result of obtaining a worklet agent given outsideSettings. Run the rest of these steps in that agent.
	- Let |operationMap| be this SharedStorageWorkletGlobalScope's operation map ...

To ensure the ordering of DB operations, we can define a parallel queue just for the DB. e.g. <dfn for="browsing context">shared storage database parallel queue</dfn>. And mention that it is associated with each browsing context and that the worklet scope will be associated with the same queue. And whenever you need to access the DB, "Enqueue the following steps on |context|'s [=browsing context/shared storage database parallel queue=]".

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+
+  The {{SharedStorage}} interface runs its methods <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel">in parallel</a>. To avoid race conditions within a single context, each context that calls into the interface will have its own <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> for handling operations.
+
+  In particular, the {{Window}}'s {{SharedStorage}} operations for a given frame will use a separate <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> from that used by any {{SharedStorageWorklet}} that may be spun up by this frame.
```
xyaoinum, line 210: For things that are linked multiple times, it's better to define in the <pre class="anchors"> section  and reference the terms via [=XXX=]. Example: https://github.com/patcg-individual-drafts/topics/blob/af873d0cfa07a174b5c3e980f23d9114b1188d07/spec.bs

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+                1. Let |operation_map| be the operation map in the associated {{SharedStorageWorkletGlobalScope}}.
+                1. Let |operation| be the operation registered to |name| in |operation_map|, if |operation| exists.
+                1. If |operation| exists:
+                    1. If |options| contains |data|, let |operation_result| be the result of invoking |operation| with arguments |url_list| and |data|.
+                    1. Otherwise, let |operation_result| be the result of invoking |operation| with argument |url_list|.
+                    1. If there are no errors:
+                        1. Set |index| to the result of casting |operation_result| to an {{unsigned long}}. If this throws an exception, catch it and set |index| back to 0.
+                        1. If |index| > |url_list|.[=list/size=], then set |index| back to 0.
+            1. Resolve |index_promise| with |index|.
+        1. When |index_promise| is resolved, let |result_index| be its numerical value.
```
xyaoinum, line 269: This sounds like there's a wait. But in reality (as in the current implementation), the JS should start allowing other operation to execute. So I think we need to queue a task, which runs the steps:
- Wait until |index_promise| is resolved.
- ....

I'm not sure if my suggestion is ideal. So perhaps leave the comment as is and wait for an spec expert.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+  <div algorithm>
+    The <dfn method for="SharedStorage">set(|key|, |value|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
+        1. If |key| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |value| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is empty, create |options| with "|ignoreIfPresent|" set to false.
+        1. Perform a database query to determine whether an |entry| for |key| already exists in the database for this context's origin.
```
xyaoinum, line 288: More rigorously define the database types and accessors. Ideally we probably should follow the terms similar to indexed db specification:
https://www.w3.org/TR/IndexedDB/#object-store
https://www.w3.org/TR/IndexedDB/#object-store-storage-operation

But as a start / for simplicity, we can model the storage as a key-value type. And define it as an user agent associated state. And perhaps add a note that "the browser should use a database backend for accessing the storage."

Each [=user agent=] has an associated <dfn for="user agent">shared storage area</dfn>, it is a map.

(When accessing it)
If user agent's [=user agent/shared storage area=] does not contain |origin|:
	- Set user agent's [=user agent/shared storage area=][|origin|] to an empty map.

Let XXX be agent's [=user agent/shared storage area=][origin],
(access the value for a key, etc…)

xyaoinum (commented):
Left some comments. General suggestion is that things should be defined/referenced more religiously (especially the database access).

wanderview:
Can you please define a bikeshed workflow for the repo so that the pull request will get a compiled preview of the spec?  Here is an example:

https://github.com/WICG/urlpattern/blob/main/.github/workflows/build.yml

I think you would probably need to add the workflow in a separate commit and then update this PR in order to get it to trigger.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+  };
+</xmp>
+
+  ## Registering a Worklet ## {#register-worklet}
+
+  <div algorithm>
+    The <dfn method for="SharedStorageWorkletGlobalScope">register(|name|, |operationCtor|)</dfn> method steps are:
+
+    1. If |name| is missing or empty, throw a {{TypeError}}.
+    1. Let |operation_map| be the internal operation <a href="https://infra.spec.whatwg.org/#ordered-map">map</a> for this scope.
```
wanderview, line 131: Its preferred to write a cross reference like `[=map=]` instead of an `<a>` tag like this.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+    1. If |name| already exists as a key in |operation_map|, throw a {{TypeError}}.
```
wanderview, line 132: Please write this like:

```
If |operation_map| [=map/contains=] an entry with key |name|, then throw a {{TypeError}}.
```

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+    1. If |operationCtor| is missing or not a constructor, throw a {{TypeError}}.
```
wanderview, line 133: @domenic What is the correct way to test if a variable contains a javascript constructor?  Similarly for accessing the prototype, etc, below.  Thanks,

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+    1. Let |prototype| be the class prototype for |operationCtor|.
+    1. If |prototype| is not an object, throw a {{TypeError}}.
+    1. Let |run| be the "run" method in |prototype|.
+    1. If |run| is missing or not a function, throw a {{TypeError}}.
+    1. <a href="https://infra.spec.whatwg.org/#map-set">Set</a> the value of |name| to |run| in |operation_map|.
```
wanderview, line 138: You can use `[=map/set=]` instead of writing the link yourself.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+</style>
 
 Introduction {#intro}
 =====================
 <em>This section is not normative.</em>
 
-This document introduces a new primitive.
+In order to prevent cross-site user tracking, browsers are partitioning all forms of storage by top-frame site. But, there are many legitimate use cases currently relying on unpartitioned storage.
+
+This document introduces a new storage API that is intentionally not partitioned by top-frame site (though still partitioned by context origin), in order to serve a number of the use cases needing unpartitioned storage. To limit cross-site reidentification of users, data in Shared Storage may only be read in a restricted environment that has carefully constructed output gates. Over time, we hope to design and add additional gates.
```
wanderview, line 30: If I understand correctly this will require a worklet with a non-opaque origin in order to access the correct shared storage bits.  AFAIK worklets have always had an opaque origin up until now.

(I know at least related aspect of this is that worklet script requests are not treated as non-subresource requests when being intercepted by a service worker.  This was previously ok because they were opaque origin globals and therefore couldn't interact with the SW anyway.  But a non-opaque origin worklet would change this.  This may just be a purity nuisance for me and not a real problem, though.)

@domenic Are you aware of any other concerns with creating non-opaque worklets?  Why are other worklets opaque in the first place?

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+                     optional SharedStorageSetMethodOptions options = {});
+
+    Promise<any> append(DOMString key,
+                        DOMString value);
+
+    Promise<any> delete(DOMString key);
+
+    Promise<any> clear();
+
+    [Exposed=(SharedStorageWorklet)] Promise<DOMString> get(DOMString key);
```
wanderview, line 176: I wonder if it would make sense to have an interface hierarchy here.  For example, WorkletSharedStorage and WindowSharedStorage would extend a base SharedStorage.  You can then set the exposed attributes on the interfaces themselves instead of on a per-method basis.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
-  boolean ignoreIfPresent;
-};
+    [Exposed=(Window)] readonly attribute SharedStorageWorklet worklet;
+  };
 
-dictionary SharedStorageRunOperationMethodOptions {
-  object data;
-};
+  dictionary SharedStorageSetMethodOptions {
+    boolean ignoreIfPresent;
```
wanderview, line 190: I think its preferred to provide a default value for dictionary attributes.

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
-  object reportingMetadata;
-};
+  dictionary SharedStorageRunOperationMethodOptions {
+    object data;
+  };
+
+  enum IteratorMode { "key", "keyValue" };
+
+  [Exposed=(SharedStorageWorklet)]
+  interface SharedStorageIterator {
```
wanderview, line 200: This is interesting.  Are we using an iterator based on the idea that there can be too many keys/entries to load in a single invocation?  I haven't read fully, but we will need to define how these iterators work if storage is modified underneath them.

Also, it would probably be preferred to use the webidl async iterator feature.  For example, see the `async iterable<any>` line in the ReadableStream spec:

https://streams.spec.whatwg.org/#readablestream

spec.bs
```diff
@@ -6,57 +6,417 @@ Status: CG-DRAFT
...
+
+  When an algorithm before asks for steps to be enqueued, these steps should be enqueued on the <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue">parallel queue</a> of the calling context, unless otherwise specified.
+
+  ## Run Operation Methods ## {#run-op}
+
+  <div algorithm>
+    The <dfn method for="SharedStorage">run(|name|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new promise.
+    1. <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#enqueue-the-following-steps">Enqueue the following steps</a>:
```
wanderview, line 220: I think its prefered to write something more like `Run the following steps [=in parallel=]:`.

spec.bs
wanderview: I believe you have to use the `[=upon fulfillment=]` hook here to take action when a promise resolves:

https://webidl.spec.whatwg.org/#upon-fulfillment

Something like `[=Upon fulfillment=] of |index_promise| perform the following steps:`.

spec.bs
wanderview: You should probably hook this into the storage spec data structures:

https://storage.spec.whatwg.org/#model

I believe the storage bucket contains a map from storage type strings (like "indexeddb") to storage bottles.

So you would monkey patch in a new registered storage type for "sharedstorage" or something, then put your map of data inside the resulting storage bottle you get.

wanderview (changes requested):
I did a partial review, but ran out of time.  This looks like a good start, but I think we need to resolve some questions and details.  Thanks!

spec.bs

spec.bs

spec.bs

spec.bs

spec.bs

spec.bs
pythagoraskitty: Updated. Let me know what you think.

spec.bs
pythagoraskitty: Apologies for the ambiguity. The worklet will have an opaque origin.  It's the origin of the worklet's creator that I was referring to.  Clarified in the spec.

spec.bs

spec.bs

spec.bs

spec.bs

spec.bs
pythagoraskitty: Ok, I think that I did this, more or less.  But please let me know if you want me to give every definition a "for" field (which seems a bit excessive to do everywhere).

pythagoraskitty:
xyaoinum@ and wanderview@, could you please take another look?  Thanks.

wanderview:
> xyaoinum@ and wanderview@, could you please take another look? Thanks.

Thanks!  I'll take a look later this week.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+
+  The {{SharedStorageWorkletGlobalScope}}'s [=module map=]'s [=module scripts=] should each define and {{register}} one or more {{SharedStorageOperation}}s. 
+
+  Each {{SharedStorageWorkletGlobalScope}} also has an associated <dfn for=SharedStorageWorkletGlobalScope>operation map</dfn>, which is an [=map=], initially empty, of [=strings=] (denoting operation names) to [=functions=].
+
+  Currently each {{SharedStorageOperation}} registered via {{SharedStorageWorkletGlobalScope/register()}} must be one of the following two types:
+  * {{SharedStorageRunOperation}}
+  * {{SharedStorageSelectURLOperation}}
+
+  A {{SharedStorageRunOperation}} is an {{SharedStorageOperation}} that has no JavaScript return value, and hence is especially suited for calling into the [=private aggregation=] service.
```
wanderview, line 243: What does it mean to have "no JavaScript return value"?  Maybe it would be better to say the return value is ignored and most of these operations are expected to return undefined.

I might also phrase the second half of this sentence as a separate sentence:

"The {{SharedStorageRunOperation}} is designed to work with output gates that do not need a return value, like the [=private aggregation=] service."

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+
+  A {{SharedStorageSelectURLOperation}} is an {{SharedStorageOperation}} that takes in a [=list=] of {{SharedStorageUrlWithMetadata}}s (i.e. [=dictionaries=] containing [=strings=] representing {{URL}}s each wrapped with metadata) and returns a {{long}} integer index specifying which of these {{URL}}s should be selected.
```
wanderview, line 245: This is not quite right.  It does not "return a {{long}}" because it returns a promise instead.  Maybe rephrase to say the inputs are a list, it performs an async operation, and then returns a promise to a long.

spec.bs
wanderview: Hmm, it does seem that the returned promise might be relevant as a way to signal completion of the operation.  Its just the fulfilled promise value is ignored.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+      USVString url;
+      object reportingMetadata;
+    };
+  </xmp>
+
+  ## Registering a Worklet ## {#register-worklet}
+  <div algorithm>
+    For the <dfn method for="SharedStorageWorkletGlobalScope">register(|name|, |operationCtor|)</dfn> method, run the following steps [=in parallel=]:
+
+    1. If |name| is missing or empty, throw a {{TypeError}}.
```
wanderview, line 269: You need to perform these steps before entering the `[=in parallel=]` block.  You can't throw any exceptions or access javascript state within the parallel code.

So move this up and then have a step that says "Run the following steps [=in parallel=]:", etc.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+    1. Let |operationMap| be this {{SharedStorageWorkletGlobalScope}}'s [=SharedStorageWorkletGlobalScope/operation map=].
+    1. If |operationMap| [=map/contains=] an [=map/entry=] with [=map/key=] |name|, throw a {{TypeError}}.
+    1. If |operationCtor| is missing, throw a {{TypeError}}.
+    1. Let |isConstructor| be the result of running [=IsConstructor()=] with |operationCtor|.
+    1. If |isConstructor| is false, throw a {{TypeError}}.
+    1. Let |prototype| be the result of running |operationCtor|'s [=object/[[GetPrototypeOf]]()=] method.
+    1. If |prototype| is not an [=object=], throw a {{TypeError}}.
+    1. Let |run| be the result of running [=GetMethod()=] with |prototype| and "`run`".
+    1. If |run| is undefined, throw a {{TypeError}}.
+    1. [=map/Set=] the value of |operationMap|[|name|] to |run|.
```
wanderview, line 279: Why do these steps need to run in parallel?  It seems like all of this can run synchronously or is there something potentially long/async here?  Also, if something runs in parallel we usually return a promise so the caller can know when its complete.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+  </div>
+
+The [=Shared Storage Database=] {#database}
+===========================================
+A [=user agent=] holds a <dfn>shared storage database</dfn>, which consists of a [=shared storage shed=] and provides methods to [=shared storage database/store an entry in the database|store=], [=shared storage database/retrieve an entry from the database|retrieve=], [=shared storage database/delete an entry from the database|delete=], [=shared storage database/clear all entries in the database|clear=], and [=shared storage database/purge expired entries from the database|purge expired=] data, and additional methods as below. The data in the [=shared storage database|database=] take the form of [=shared storage database/entry|entries=].
```
wanderview, line 284: I bet we might get a later request to merge `shared storage shed` and `storage shed`.  But what you have here seems fine for now given that a storage shed "is a map" and not a container holding a map.  No action required here.

wanderview, line 284: Here you say the user agent holds the database and the database holds the shed, but below you say the user agent holds the shed directly.  Which one is correct?

I think probably we want the user agent to hold the shed directly and then put the shared storage database inside one of the storage spec layers.  I think the storage bucket layer makes the most sense.  Its the layer where other storage APIs are defined, etc.  This would mean defining a new storage endpoint identifier:

https://storage.spec.whatwg.org/#registered-storage-endpoints

Which it looks like you do below with the string "shared".

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+  </div>
+
+
+
+  ## Monkey Patch for the [=Storage Model=] ## {#monkey-patch}
+  After adding a new [=storage type=] "`shared`" to the [=Storage Model=], this standard will [=storage endpoint/register=] a [=storage endpoint=] of [=storage type|type=] "`shared`" with [=storage identifier=] "`sharedStorage`" and [=storage endpoint/quota=] `5`<sup>`4`</sup> `*` `2`<sup>`16`</sup> bytes (i.e. 39.0625 mebibytes).
+
+  <span class=todo>This [=storage endpoint/quota=] is calculated from the current implementation. Consider bringing the current implementation in line with the spec for [=storage endpoints=] "`localStorage`" and "`sessionStorage`", i.e. `5 * 2`<sup>`20`</sup> bytes. For example, decreasing the per-origin entry limit from 10,000 to 1,280 would achieve this.</span>
+
+  A [=user agent=] holds a [=shared storage shed=] for [=storage endpoints=] of [=storage type|type=] "`shared`".
```
wanderview, line 396: Maybe also mention that the storage shelves inside of shared storage shed always have a key of "default" for now.  (Because we don't support the buckets API yet.)

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+
+  A <dfn>shared storage shed</dfn> is a [=map=] of [=/origins=] to to [=storage shelf|storage shelves=]. It is initially empty.
+
+  Note: Unlike [=storage sheds=], whose keys are [=storage keys=], [=shared storage sheds=] use [=/origins=] as keys directly. [=Shared storage=] will be intentionally excluded from [=client-side storage partitioning=].
+
+  A [=user agent=]'s [=shared storage shed=] holds all <dfn>shared storage</dfn> data.
+
+  A <dfn>shared storage shelf</dfn> is a [=storage shelf=] in a [=shared storage shed=].
```
wanderview, line 404: Why do we need a new shelf definition?  It seems the existing storage shelf should work?   It will just only have a "default" key in its bucket map because we don't support using shared storage in different storage buckets right now.

wanderview (changes requested):
I didn't make it all the way through, but wanted to share comments so far.  Thank you for integrating with storage spec.  I think maybe we just need to clarify a few things there.  In particular where the shared storage database lives in the data structure.

I'll keep looking more later this week.  Thanks!

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+[=User agents=] may specify a <dfn>default entry lifetime</dfn>, the default duration between when an [=shared storage database/entry=] is [=shared storage database/store an entry in the database|stored=] and when it expires.
+
+A <dfn>shared storage database queue</dfn> for a [=/browsing context=] is a [=parallel queue=] on which the each of the [=shared storage database=]'s methods run when calls to them are initiated from that [=/browsing context=].
+
+  ## The [=Shared Storage Database|Database=] Algorithms ## {#database-algorithms}
+
+  <div algorithm>
+    To <dfn for="shared storage database">store an entry in the database</dfn>, given a [=/browsing context=] |context|, a [=entry/key=] |key|, and a [=value struct/value=] |value|, [=enqueue the following steps=] on |context|'s associated [=shared storage database queue=]:
+
+    1. If |key| is not a [=string=] or exceeds the [=key/maximum length=], return failure.
```
xyaoinum, line 307: Is "return failure" a standard term? I think this algorithm should return either true or false (ditto for others), and in the caller (e.g. worklet), check the result, and resolve or reject the promise accordingly.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+
+  dictionary SharedStorageRunOperationMethodOptions {
+    object data;
+  };
+</xmp>
+
+  ### Window Setter/Deleter Methods ### {#window-setter}
+
+  <div algorithm>
+    For the <dfn method for="WindowSharedStorage">set(|key|, |value|, |options|)</dfn> method, run the following steps [=in parallel=]:
```
xyaoinum, line 501: The algorithm for IDL methods are usually: "The ... method steps are:".
Also, there's no need to mention "in parallel" in any of these steps:
- "Enqueue the following steps" should be sufficient.
- Many steps (e.g. the param validation part) can be put before "Enqueue the following steps". Only the DB access that can take a longer time should be put inside the "Enqueue".

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+                    1.  Let |operation| be |operationMap|[|name|].
+                    1. Let |argumentsList| be a new [=list=].
+                    1. [=list/Append=] |urlList| to |argumentsList|.
+                    1. If |options| [=map/contains=] |data|, [=list/Append=] |data| to |argumentsList|.
+                    1. If |options| contains |data|, let |operationResult| be the result of running [=Call=] on |operation| with |argumentsList|.
+                    1. If there are no errors:
+                        1. Set |index| to the result of casting |operationResult| to an {{unsigned long}}. If this throws an exception, catch it and set |index| back to [=default index=].
+                        1. If |index| is greater than |urlList|.[=list/size=], then set |index| back to [=default index=].
+            1. Resolve |indexPromise| with |index|.
+        1. [=Upon fulfillment=] of |indexPromise|, perform the following steps:
```
xyaoinum, line 706: Should there be a [=Upon rejection=] scenario as well?

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+  };
 </xmp>
+
+  ### Worklet Setter/Deleter Methods ### {#worklet-setter}
+
+  <div algorithm>
+    For the <dfn method for="WorkletSharedStorage">set(|key|, |value|, |options|)</dfn> method, run the following steps [=in parallel=]:
+
+    1. Let |promise| be a new promise.
+    1. Let |context| be {{WorkletSharedStorage}}'s {{SharedStorageWorkletGlobalScope}}'s [[=outside settings=]'s [=target browsing context=].
```
xyaoinum, line 741: There's redundant  "[".

spec.bs
pythagoraskitty: Yes, it's what the Storage Standard uses. See https://storage.spec.whatwg.org/#model and e.g. https://storage.spec.whatwg.org/#obtain-a-storage-shelf.

I will continue to use "failure" for the storage monkey patch, but changed to true/false or null/undefined elsewhere as applicable.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+  top: -0.25em;
+}
+</style>
+
 Introduction {#intro}
 =====================
 <em>This section is not normative.</em>
 
-This document introduces a new primitive.
+In order to prevent cross-site user tracking, browsers are partitioning all forms of storage by top-frame site; see [=Client-Side Storage Partitioning=]. But, there are many legitimate use cases currently relying on unpartitioned storage.
```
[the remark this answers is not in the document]
pythagoraskitty: Linked to examples in the explainer.

spec.bs
```diff
@@ -6,57 +6,882 @@ Status: CG-DRAFT
...
+    [Exposed=SharedStorageWorklet]
+    interface SharedStorageOperation {
+    };
+  </xmp>
+
+  Each {{SharedStorageWorkletGlobalScope}} has an associated [=environment settings object=] <dfn for=SharedStorageWorkletGlobalScope>outside settings</dfn>, which is the associated {{SharedStorageWorklet}}'s [=relevant settings object=].
+
+  The {{SharedStorageWorkletGlobalScope}}'s [=module map=]'s [=module scripts=] should each define and {{register}} one or more {{SharedStorageOperation}}s. 
+
+  Each {{SharedStorageWorkletGlobalScope}} also has an associated <dfn for=SharedStorageWorkletGlobalScope>operation map</dfn>, which is an [=map=], initially empty, of [=strings=] (denoting operation names) to [=functions=].
```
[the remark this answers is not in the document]

spec.bs

spec.bs
pythagoraskitty: Removed "[=in parallel=]" per your comment below and xyaoinum@'s comment.

spec.bs
pythagoraskitty: I believe I addressed this now.

spec.bs
pythagoraskitty: Noted, thanks.

pythagoraskitty: Changed to be somewhat similar to  https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute

spec.bs
pythagoraskitty: Rephrased.

spec.bs
pythagoraskitty: I think that I've addressed now.

spec.bs
pythagoraskitty: Added.

spec.bs

spec.bs

spec.bs
pythagoraskitty: Changed.

pythagoraskitty:
wanderview@ and xyaoinum@, please take another look when you get a chance.

Thank you!

spec.bs
```diff
@@ -6,57 +6,924 @@ Status: CG-DRAFT
...
+
+  This standard will add a new [=storage type=] "`shared`" to the [=Storage Model=].
+
+  A [=user agent=] holds a [=shared storage shed=] for [=storage endpoints=] of [=storage type|type=] "`shared`".
+
+  This standard will also [=storage endpoint/register=] a [=storage endpoint=] of [=storage type|type=] "`shared`" with [=storage identifier=] "`sharedStorage`" and [=storage endpoint/quota=] `5`<sup>`4`</sup> `*` `2`<sup>`16`</sup> bytes (i.e. 39.0625 mebibytes).
+
+  <span class=todo>This [=storage endpoint/quota=] is calculated from the current implementation. Consider bringing the current implementation in line with the spec for [=storage endpoints=] "`localStorage`" and "`sessionStorage`", i.e. `5 * 2`<sup>`20`</sup> bytes. For example, decreasing the per-origin entry limit from 10,000 to 1,280 would achieve this.</span>
+
+  A <dfn>shared storage shed</dfn> is a [=map=] of [=/origins=] to to [=storage shelf|storage shelves=]. It is initially empty.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -6,57 +6,924 @@ Status: CG-DRAFT
...
+  ### Window Setter/Deleter Methods ### {#window-setter}
+
+  <div algorithm>
+    The <dfn method for="WindowSharedStorage">set(|key|, |value|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new [=promise=].
+    1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
+    1. [=Assert=] that |context| is not null.
+    1. [=Enqueue the following steps=] on |context|'s associated [=shared storage database/shared storage database queue=]:
+        1. If |key| is not a {{DOMString}} or exceeds the [=key/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -6,57 +6,924 @@ Status: CG-DRAFT
...
+  </div>
+
+  <div algorithm>
+    The <dfn method for="WindowSharedStorage">append(|key|, |value|)</dfn> method steps are:
+
+    1. Let |promise| be a new [=promise=].
+    1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
+    1. [=Assert=] that |context| is not null.
+    1. [=Enqueue the following steps=] on |context|'s associated [=shared storage database/shared storage database queue=]:
+        1. If |key| is not a {{DOMString}} or exceeds the [=key/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -6,57 +6,924 @@ Status: CG-DRAFT
...
+        1. If |value| is not a {{DOMString}} or exceeds the [=value/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+        1. If |options| is empty, create |options|.
+        1. If |options|["`ignoreIfPresent`"] is true:
+            1. Let |currentValue| be the result of running [=shared storage database/retrieve an entry from the database=] with |context| and |key|.
+            1. If |currentValue| is not null and |currentValue| is not undefined, resolve |promise| as undefined.
+            1. If |currentValue| is null, reject |promise|.
+            1. Otherwise, let |result| be the result of running [=shared storage database/store an entry in the database=] with |context|, |key|, and |value|.
```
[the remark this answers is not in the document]
pythagoraskitty: Done.

spec.bs
```diff
@@ -6,57 +6,924 @@ Status: CG-DRAFT
...
+
+  Each <dfn for="shared storage database">entry</dfn> consists of a [=entry/key=] and a [=entry/value struct=].
+
+  An [=shared storage database/entry=]'s <dfn for=entry>key</dfn> is a [=string=].
+
+  [=User agents=] may specify the <dfn for=key>maximum length</dfn> of a [=entry/key=].
+
+  Since [=entry/keys=] are used to organize and efficiently retrieve [=shared storage database/entry|entries=], [=entry/keys=] must appear at most once in any given [=shared storage database=].
+
+  An [=shared storage database/entry=]'s <dfn for=entry>value struct</dfn> is a [=struct=] composed of [=string=] <dfn for="value struct">value</dfn> and timestamp <dfn for="value struct">last updated</dfn>.
```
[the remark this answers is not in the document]
pythagoraskitty: Done, thanks!

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+
+  [=User agents=] may specify the <dfn for=key>maximum length</dfn> of a [=entry/key=].
+
+  Since [=entry/keys=] are used to organize and efficiently retrieve [=shared storage database/entry|entries=], [=entry/keys=] must appear at most once in any given [=shared storage database=].
+
+  An [=shared storage database/entry=]'s <dfn for=entry>value struct</dfn> is a [=struct=] composed of [=string=] <dfn for="value struct">value</dfn> and {{DOMHighResTimeStamp}} <dfn for="value struct">last updated</dfn> (from the [=Unix Epoch=]).
+
+  Note: Some Shared Storage implementations may use SQLite implementations that store timestamps as 64-bit unsigned integers rather than doubles, thereby making it necessary to use legacy {{EpochTimeStamp}}s instead of {{DOMHighResTimeStamp}}s.
+
+  <span class=todo>In Chrome, for the reason given in the note above, we use {{EpochTimeStamp}}s, which are discouraged. How can we reconcile this?</span>
```
wanderview, line 368: Can you explain this a bit more?  Surely sqlite implementations can store double values.  What prevents us from accurately representing a DOMHighResTimeStamp here?  Can we not convert whatever our internal representation is into a DOMHighResTimeStamp when its exposed to script?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+        1. Let |databaseMap| be the result of running [=obtain a shared storage bottle map=] for |environment|.
+        1. If |databaseMap| is failure, then return null.
+        1. Let |valueStruct| be a new [=entry/value struct=].
+        1. Set |valueStruct|'s [=value struct/value=] to |value|.
+        1. Let |unsafeMoment| be the [=wall clock=]'s [=wall clock/unsafe current time=].
+        1. Let |moment| be the result of running [=coarsen time=] algorithm given |unsafeMoment| and [=wall clock=] as input.
+        1. Let |fromUnixEpochTime| be the [=duration from=] the [=Unix epoch=] to |moment|.
+        1. Set |valueStruct|'s [=value struct/last updated=] to |fromUnixEpochTime|.
+        1. [=map/Set=] |databaseMap|[|key|] to |valueStruct|.
+        1. Return true.
```
wanderview, line 392: Unfortunately you cannot return a value like this from an enqueued set of steps.  The algorithm synchronously returns undefined before these steps are executed.

There are a couple ways to solve this:

1.  You can make these internal algorithms return promises.  If the promises are always directly returned to javascript through the API layer, then that may be fine.  If you want to use these values inside other algorithms, though, it might get a bit ugly.
2. You can make these internal algorithms synchronous and require that they are only called via the database queue.  The outer API method algorithms then would be responsible for enqueuing the steps that call the internal algorithm.  If you do this you will need to pass in any information needed from the current context; e.g. the environment, etc.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+
+    1. If |key| is not a [=string=] or |key|'s [=string/length=] exceeds the [=key/maximum length=], return null.
+    1. Let |environment| be |context|'s [=active window=]'s [=relevant settings object=].
+    1. [=Enqueue the following steps=] on |context|'s associated [=shared storage database/shared storage database queue=]:
+        1. Let |databaseMap| be the result of running [=obtain a shared storage bottle map=] for |environment|.
+        1. If |databaseMap| is failure, then return null.
+        1. If |databaseMap| does not [=map/contain=] |key|, return undefined.
+        1. Let |valueStruct| be the result of running [=map/Get=] on |databaseMap| with |key|.
+        1. Let |isExpired| be the result of running [=shared storage database/determine whether an entry is expired=] with |valueStruct|.
+        1. If |isExpired| is true, return undefined.
```
wanderview, line 406: I think its preferred to combine things like these two steps into a single step.  Instead of creating the `isExpired` variable and only using it in one step, instead write something like: `If the result of running [=shared storage database/determine whether an entry is expired=] given |valueStruct| is true, then return undefined.`

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+        1. If |databaseMap| is failure, then return undefined.
+        1. Let |count| be |databaseMap|'s [=map/size=].
+        1. Return |count|.
+  </div>
 
-[Exposed=(Window)]
-interface SharedStorage {
-  Promise<any> set(DOMString key, DOMString value, optional SharedStorageSetMethodOptions options = {});
+  <div algorithm>
+    To <dfn for="shared storage database">purge expired entries from the database</dfn>, given a [=/browsing context=] |context|, run the following steps:
```
wanderview, line 457: What calls this algorithm?  Should we have something that says the user agent has a timer that periodically calls this algorithm?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. Let |proxyMap| be a new [=storage proxy map=] whose [=backing map=] is |bottle|'s [=map=].
+    1. [=set/Append=] |proxyMap| to |bottle|'s [=proxy map reference set=].
+    1. Return |proxyMap|.
+  </div>
+
+  ## The [=Shared Storage Database=] ## {#database}
+
+  A [=/browsing context=] has an associated <dfn>shared storage database</dfn>, which provides methods to [=shared storage database/store an entry in the database|store=], [=shared storage database/retrieve an entry from the database|retrieve=], [=shared storage database/delete an entry from the database|delete=], [=shared storage database/clear all entries in the database|clear=], and [=shared storage database/purge expired entries from the database|purge expired=] data, and additional methods as below. The data in the [=shared storage database|database=] take the form of [=shared storage database/entry|entries=].
+
+  Each [=shared storage database=] has a <dfn for="shared storage database">shared storage database queue</dfn>, which is a [=parallel queue=] on which the each of the [=shared storage database=]'s methods run when calls to them are initiated from that [=/browsing context=].
```
wanderview, line 354: I think you need to say the queue is the resulting of starting the parallel queue:

https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+
+  ### Window Setter/Deleter Methods ### {#window-setter}
+
+  <div algorithm>
+    The <dfn method for="WindowSharedStorage">set(|key|, |value|, |options|)</dfn> method steps are:
+
+    1. Let |promise| be a new [=promise=].
+    1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
+    1. [=Assert=] that |context| is not null.
+    1. If |key|'s [=string/length=] exceeds the [=key/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
```
wanderview, line 540: I think the current guidance is not to use fine-grained DOMExceptions like DataError.  Instead just use a TypeError.  Here and elsewhere.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. If |value|'s [=string/length=] exceeds the [=value/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
```
wanderview, line 542: What does it mean for options to be invalid?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. If |options| is empty, create |options|.
```
wanderview, line 543: Options cannot be empty since the idl has a default value to an empty struct.  So I don't think you need this step.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+        1. Let |result| be null.
+        1. Let |currentValue| be the result of running [=shared storage database/retrieve an entry from the database=] with |context| and |key|.
+        1. If |currentValue| is not failure:
+            1. Let |list| be a new [=list=].
+            1. [=list/Append=] |currentValue| to |list|.
+            1. [=list/Append=] |value| to |list|.
+            1. Let |newValue| be the result of running [=string/concatenate=] on |list|.
+            1. Let |result| be the result of running [=shared storage database/store an entry in the database=] with |context|, |key|, and |newValue|.
+        1. Otherwise, let |result| be the result of running [=shared storage database/store an entry in the database=] with |context|, |key|, and |value|.
+        1. If |result| is false, reject |promise|.
```
wanderview, line 575: Promises should be rejected with values like TypeError.  Something like `reject |promise| with a {{TypeError}}`.  Here and elsewhere.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+        1. Otherwise, resolve |promise| as undefined.
```
wanderview, line 576: nit:  I think we prefer "with" instead of "as" here.  So `resolve |promise| with undefined`.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. Let |epochLength| be [=budget lifetime=].
+    1. Let |unsafeMoment| be the [=wall clock=]'s [=wall clock/unsafe current time=].
+    1. Let |moment| be the result of running [=coarsen time=] algorithm given |unsafeMoment| and [=wall clock=] as input.
+    1. Let |fromUnixEpochTime| be the [=duration from=] the [=Unix epoch=] to |moment|.
+    1. Let |threshold| be |fromUnixEpochTime| minus |epochLength|.
+    1. Let |debitSum| be 0.
+    1. For each [=list/item=] |bitDebit| in |ledger|, do the following steps:
+        1. Let |timestamp| be |bitDebit|'s [=bit debit/timestamp=].
+        1. Let |debit| be |bitDebit|'s [=bit debit/bits=].
+        1. If |timestamp| is greater than or equal to |threshold|, then increment |debitSum| by |debit|.
```
wanderview, line 663: Should we somehow cleanup expired entries in the ledger?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
-  object data;
-};
+    1. If |origin| is [=opaque origin|opaque=], return undefined.
+    1. Let |maxBits| be the value of the [=user agent=]'s [=entropy allowance=].
+    1. If the [=user agent=]'s [=shared storage budget table=] does not [=map/contain=] |origin|, then return |maxBits|.
+    1. Otherwise, let |ledger| be [=user agent=]'s [=shared storage budget table=][|origin|].
+    1. Let |epochLength| be [=budget lifetime=].
+    1. Let |unsafeMoment| be the [=wall clock=]'s [=wall clock/unsafe current time=].
+    1. Let |moment| be the result of running [=coarsen time=] algorithm given |unsafeMoment| and [=wall clock=] as input.
+    1. Let |fromUnixEpochTime| be the [=duration from=] the [=Unix epoch=] to |moment|.
```
wanderview, line 657: Do we really need to convert into the epoch representation here?  Since this is all internal data it seems like we could just work with moments and durations?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+                1. Otherwise, [=call=] |operation| without any arguments list.
+        1. Resolve |promise| as undefined.
+    1. Return |promise|.
+  </div>
+
+  <div algorithm>
+    The <dfn method for="WindowSharedStorage">selectURL(|name|, |urls|, |options|)</dfn> method steps are:
+
+    1. Let |configPromise| be a new [=promise=].
+    1. If |urls| is empty or exceeds the maximum allowed length, reject |configPromise|.
```
wanderview, line 721: I think you want to stop the method at this point.  So you probably want to write something like `return a [=promise rejected=] with a {{TypeError}}`.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. Let |urlList| be an empty {{list}}.
+    1. For each |urlWithMetadata| in |urls|:
+        1. If |urlWithMetadata| has no field "`url`", reject |configPromise|.
+        1. Otherwise, let |urlString| be |urlWithMetadata|["`url`"].
+        1. Let |url| be the result of running a [=URL parser=] on |urlString|.
+        1. If |url| is not a valid {{URL}}, reject |configPromise|.
+        1. Otherwise, [=list/append=] |urlString| to |urlList|.
```
wanderview, line 728: Why do we parse the URL if we are going to use the original string?  Should we instead be serializing the parsed url back to a string so we get any normalization that occurred?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. Run the following steps in |agent|:
+        1. Let |index| be [=default index=].
+        1. If {{WindowSharedStorage/worklet}}'s [=module map=] is not [=map/empty=]:
+            1. Let |operationMap| be the associated {{SharedStorageWorkletGlobalScope}}'s [=SharedStorageWorkletGlobalScope/operation map=].
+            1. If |operationMap| [=map/contains=] |name|:
+                1.  Let |operation| be |operationMap|[|name|].
+                1. Let |argumentsList| be a new [=list=].
+                1. [=list/Append=] |urlList| to |argumentsList|.
+                1. If |options| [=map/contains=] |data|, [=list/Append=] |data| to |argumentsList|.
+                1. If |options| contains |data|, let |operationResult| be the result of running [=Call=] on |operation| with |argumentsList|.
```
wanderview, line 747: Is this the same if-statement as the line above?  Should we instead have one check and then two sub-steps?

wanderview (changes requested):
Thanks for all the changes.  Here is another incremental review.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+  The {{SharedStorageWorklet}}'s [=worklet global scope type=] is {{SharedStorageWorkletGlobalScope}}. 
+
+  <xmp class='idl'>
+    [Exposed=SharedStorageWorklet, Global=SharedStorageWorklet]
+    interface SharedStorageWorkletGlobalScope : WorkletGlobalScope {
+      undefined register(DOMString name,
+                         SharedStorageOperationConstructor operationCtor);
+    };
+
+    callback SharedStorageOperationConstructor = SharedStorageOperation(object options);
```
wanderview, line 228: How is the `object options` argument to the SharedStorageOperation used?  Can that be a more specific type than "object"?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
-  Promise<any> delete(DOMString key);
+  <div algorithm>
+    To <dfn for="shared storage database">determine whether an entry is expired</dfn>, given a [=entry/value struct=] |valueStruct|, run the following steps:
 
-  Promise<any> clear();
+    1. Let |lastUpdated| be |valueStruct|'s [=value struct/last updated=].
+    1. Let |lifetime| be [=user agent=]'s [=default entry lifetime=].
+    1. Let |expiration| be the sum of |lastUpdated| and |lifetime|.
+    1. Let |unsafeMoment| be the [=wall clock=]'s [=wall clock/unsafe current time=].
+    1. Let |moment| be the result of running [=coarsen time=] algorithm given |unsafeMoment| and [=wall clock=] as input.
```
wanderview, line 477: I think you can use `[=current wall time=]` instead of these two steps.  It gets the unsafe current time and automatically runs it through the coarsen time step.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. If |key|'s [=string/length=] exceeds the [=key/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |value|'s [=string/length=] exceeds the [=value/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |options| is empty, create |options|.
+    1. [=Enqueue the following steps=] on |context|'s associated [=shared storage database/shared storage database queue=]:
+        1. Let |result| be null.
+        1. If |options|["`ignoreIfPresent`"] is true:
+            1. Let |currentValue| be the result of running [=shared storage database/retrieve an entry from the database=] with |context| and |key|.
+            1. If |currentValue| is not null and |currentValue| is not undefined, resolve |promise| as undefined.
+            1. If |currentValue| is null, reject |promise|.
```
wanderview, line 549: What is the significance of the current value being `null`?  Why does this cause the operation to fail instead of setting the new value?

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+            1. Otherwise, let |result| be the result of running [=shared storage database/store an entry in the database=] with |context|, |key|, and |value|.
```
wanderview, line 550: Rather than writing the "otherwise" step twice, I think you can abort the parallel steps after rejecting the promise above.  Like `reject |promise| with {{TypeError}} and abort these steps`.  Then you can drop the "otherwise" word below and unconditionally so that the value is always set.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. While |node| is not null:
+        1. If |node| has a {{FencedFrameConfig}}:
+            1. Let |origin| be |node|'s [=active document=]'s [=document/origin=].
+            1. Let |pendingBits| be |node|'s {{FencedFrameConfig}}'s [=pending shared storage budget debit=].
+            1. If |pendingBits| is greater than 0, run the following steps:
+                1. Let |ledger| be [=user agent=]'s [=shared storage budget table=][|origin|].
+                1. Let |bitDebit| be a new [=bit debit=].
+                1. Set |bitDebit|'s [=bit debit/bits=] to |pendingBits|.
+                1. Let |unsafeMoment| be the [=wall clock=]'s [=wall clock/unsafe current time=].
+                1. Let |moment| be the result of running [=coarsen time=] algorithm given |unsafeMoment| and [=wall clock=] as input.
```
wanderview, line 686: Again, `[=current wall time=]` is the coarsened wall clock's current time.  So I think you can simply those two steps into one.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. If |key|'s [=string/length=] exceeds the [=key/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |value|'s [=string/length=] exceeds the [=value/maximum length=], reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |options| is invalid, reject |promise| with a "{{DataError}}" {{DOMException}}.
+    1. If |options| is empty, create |options|.
+    1. [=Enqueue the following steps=] on |context|'s associated [=shared storage database/shared storage database queue=]:
+        1. Let |result| be null.
+        1. If |options|["`ignoreIfPresent`"] is true:
+            1. Let |currentValue| be the result of running [=shared storage database/retrieve an entry from the database=] with |context| and |key|.
+            1. If |currentValue| is not null and |currentValue| is not undefined, resolve |promise| as undefined.
+            1. Otherwise, if |currentValue| is null, reject |promise|.
```
wanderview, line 801: Again, why does reject if the existing value is null instead of overwriting it?

wanderview, line 801: Also, this algorithm seems duplicated with the one above.  Should this be calling into the other one or a common algorithm they both use?  Is there a reason for them to be duplicated?  Are they slightly different?

wanderview (changes requested):
A few more comments as I ran short on time previously.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+===================================
+The Shared Storage API will integrate into the [=Storage Model|Storage API=] as below, via [=storage endpoint/registering=] a new [=storage endpoint=].
+
+  ## Monkey Patch for the [=Storage Model=] ## {#monkey-patch}
+
+  This standard will add a new [=storage type=] "`shared`" to the [=Storage Model=].
+
+  A [=user agent=] holds a [=shared storage shed=] for [=storage endpoints=] of [=storage type|type=] "`shared`".
+
+  This standard will also [=storage endpoint/register=] a [=storage endpoint=] of [=storage type|type=] "`shared`" with [=storage identifier=] "`sharedStorage`" and [=storage endpoint/quota=] `5`<sup>`4`</sup> `*` `2`<sup>`16`</sup> bytes (i.e. 39.0625 mebibytes).
```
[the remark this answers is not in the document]
pythagoraskitty: Yes, reporting the usage of shared storage to anyone beside the user would be a privacy leak, as it could be used as a side channel.

spec.bs
```diff
@@ -6,57 +6,940 @@ Status: CG-DRAFT
...
+    1. Let |operationMap| be this {{SharedStorageWorkletGlobalScope}}'s [=SharedStorageWorkletGlobalScope/operation map=].
+    1. If |operationMap| [=map/contains=] an [=map/entry=] with [=map/key=] |name|, throw a {{TypeError}}.
+    1. If |operationCtor| is missing, throw a {{TypeError}}.
+    1. Let |isConstructor| be the result of running [=IsConstructor()=] with |operationCtor|.
+    1. If |isConstructor| is false, throw a {{TypeError}}.
+    1. Let |prototype| be the result of running |operationCtor|'s [=object/[[GetPrototypeOf]]()=] method.
+    1. If |prototype| is not an [=object=], throw a {{TypeError}}.
+    1. Let |run| be the result of running [=GetMethod()=] with |prototype| and "`run`".
+    1. If |run| is undefined, throw a {{TypeError}}.
+    1. [=map/Set=] the value of |operationMap|[|name|] to |run|.
```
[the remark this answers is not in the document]
pythagoraskitty: No need for an IPC, correct.

spec.bs
pythagoraskitty: Here is the code reference:
https://source.chromium.org/chromium/chromium/src/+/main:sql/statement.cc;drc=c85c710dec58dfdcf2f35af647a0a5155639edd5;l=469

It would probably not be impossible to roll our own version using a double, but I don't think that's best practice.

But sure, we could convert our int to a double when it's exposed to script.

wanderview: Yea, I think we should write the spec without these kinds of implementation conversion details.  It shouldn't really mention sqlite as a possible implementation database under the hood.  If there is any conversion needed we should just do it in our code and not bring it up in the spec.

spec.bs
pythagoraskitty: Made synchronous, eliminated double checking the inputs

spec.bs

spec.bs

pythagoraskitty:
wanderview@ and xyaoinum@ could you please take another look? Thank you!

spec.bs
wanderview: Hmm, many specs use verbiage like "return failure" and then "if |result| is failure" to signify this sort of thing.  That might be more clear.

spec.bs
```diff
@@ -6,57 +6,1065 @@ Status: CG-DRAFT
...
+  <div algorithm>
+    The <dfn method for="WorkletSharedStorage">length()</dfn> method steps are:
+
+    1. Let |promise| be a new [=promise=].
+    1. Let |context| be {{WorkletSharedStorage}}'s {{SharedStorageWorkletGlobalScope}}'s [=outside settings=]'s [=target browsing context=].
+    1. Let |environment| be |context|'s [=active window=]'s [=relevant settings object=].
+    1. Let |queue| be |context|'s associated [=shared storage database/shared storage database queue=].
+    1. [=Enqueue the following steps=] on |queue|:
+        1. Let |numEntries| be the result of running [=shared storage database/count entries in the database=] with |queue| and |environment|.
+        1. If |numEntries| is undefined, reject |promise| with a {{TypeError}}.
```
wanderview, line 1002: Maybe another place you could use "is failure" instead.

wanderview (approved):
A few more nits, but overall I think this looks very solid.  I think it would be fair to fix these last few things and then land it.  Any additional changes can be done via issues/PRs.

Thanks for all your hard work on this!

spec.bs
```diff
@@ -6,57 +6,1065 @@ Status: CG-DRAFT
...
+    1. [=set/Append=] |proxyMap| to |bottle|'s [=proxy map reference set=].
+    1. Return |proxyMap|.
+  </div>
+
+  ## The [=Shared Storage Database=] ## {#database}
+
+  A [=/browsing context=] has an associated <dfn>shared storage database</dfn>, which provides methods to [=shared storage database/store an entry in the database|store=], [=shared storage database/retrieve an entry from the database|retrieve=], [=shared storage database/delete an entry from the database|delete=], [=shared storage database/clear all entries in the database|clear=], and [=shared storage database/purge expired entries from the database|purge expired=] data, and additional methods as below. The data in the [=shared storage database|database=] take the form of [=shared storage database/entry|entries=].
+
+  Each [=shared storage database=] has a <dfn for="shared storage database">shared storage database queue</dfn>, which is the result of [=start a new parallel queue|starting a new parallel queue=], on which the each of the [=shared storage database=]'s methods run when calls to them are initiated from that [=/browsing context=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -6,57 +6,1065 @@ Status: CG-DRAFT
...
-};
+    1. Let |promise| be a new [=promise=].
+    1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
+    1. [=Assert=] that |context| is not null.
+    1. If |key|'s [=string/length=] exceeds the [=key/maximum length=], return a [=promise rejected=] with a {{TypeError}}.
+    1. If |value|'s [=string/length=] exceeds the [=value/maximum length=], return a [=promise rejected=] with a {{TypeError}}.
+    1. Let |environment| be |context|'s [=active window=]'s [=relevant settings object=].
+    1. Let |queue| be |context|'s associated [=shared storage database/shared storage database queue=].
+    1. [=Enqueue the following steps=] on |queue|:
+        1. If |options|["`ignoreIfPresent`"] is true and the result of running [=shared storage database/retrieve an entry from the database=] with |queue|, |environment|, and |key| is not undefined, return [=promise resolved=] with undefined.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -6,57 +6,1065 @@ Status: CG-DRAFT
...
+
+  Inside a [=fenced frame=] with [=FenceEvent/eventType=]-[=/URL=] pairs that have been [=register reporting metadata|registered=] through {{WindowSharedStorage/selectURL()}} with {{SharedStorageUrlWithMetadata/reportingMetadata}} {{/object}}s, if [=fence.reportEvent()=] is called on a [=FenceEvent=] with a [=FenceEvent/destination=] [=list/containing=] "`shared-storage-select-url`" and that [=FenceEvent=]'s corresponding [=FenceEvent/eventType=] is triggered, then the [=FenceEvent=]'s [=FenceEvent/eventData=] will be sent as a [=beacon=] to the registered [=/URL=] for that [=FenceEvent/eventType=].
+
+  <div algorithm>
+    To <dfn>validate reporting metadata</dfn>, given an {{/object}} |reportingMetadata|, run the following steps:
+
+    1. If |reportingMetadata| is [=map/empty=], return true.
+    1. If |reportingMetadata| is not a [=dictionary=], return false.
+    1. For each <var ignore="">eventType</var> -> |urlString| of |reportingMetadata|, if the result of running [=get the canonical URL string if valid=] with |urlString| is undefined, return false.
```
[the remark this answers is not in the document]

spec.bs

xyaoinum (approved):
LGTM / agree that we can create issues/PRs to address additional changes

PR #63, opened 2023-02-16
Add API to retrieve embedder context from inside a fenced frame

pythagoraskitty:
We propose adding to the API surfaces of both Shared Storage and Fenced Frames so that a string of contextual information can be written by the fenced frames's embedder and then retrieved inside a shared storage worklet nested in the fenced frame.


pythagoraskitty:
jkarlin@ and xyaoinum@, could you please take a look? Thank you.

README.md
```diff
@@ -139,6 +139,9 @@ There have been multiple privacy proposals ([SPURFOWL](https://github.com/AdRoll
     *   Same as outside the worklet, except that the promise returned only resolves into `undefined` when the operation has completed.
 *   `sharedStorage.remainingBudget()`
     *   Returns a number indicating the remaining available privacy budget for `sharedStorage.selectURL()`, in bits.   
+*  `sharedStorage.embedderContext()`
+    *   From inside a worklet created inside a [fenced frame](https://github.com/wicg/fenced-frame/), returns a promise that resolves to a string of contextual information, if any, that the embedder had written to the [fenced frame](https://github.com/wicg/fenced-frame/)'s [FencedFrameConfig](https://wicg.github.io/fenced-frame/#fencedframeconfig) before the [fenced frame](https://github.com/wicg/fenced-frame/)'s navigation.
```
jkarlin, line 143: Do we have to return a promise? Would be nice if this could be synchronous. It seems like we could send the contextual information to the worklet on worklet creation. 

That would imply that the contextual information can't be changed (or at least, can't be read again) after worklet creation. Is that acceptable/desirable?

pythagoraskitty: In my current design there will be an IPC from the worklet process to the browser process in order to get this information.

If instead we give the information to the worklet process at creation as you suggest, there would be no need for an IPC, and then yes, the call could be synchronous.  

As it is, in my current design, the information cannot be changed after navigation to the config.  That is, only information written to the config before the config is navigated to will actually be available via this API.

pythagoraskitty: I just took a look at the code, and it should be doable to plumb the context to eligible worklets during their initialization.  Will update the PR to make the call synchronous.

jkarlin (approved):

PR #67, opened 2023-03-03
Add `keepAlive` option to explainer

pythagoraskitty:
We add the option to `keepAlive` a worklet after a `selectURL`/`run` operation. This option will default to false, and if not specified as true, the worklet will close after the operation completes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,11 +313,9 @@ register('send-report', ReportingOperation);
 
 ### Keeping a worklet alive for multiple operations
 
-We may wish to run multiple worklet operations from the same context, e.g. we might select a URL and then send one or more aggregatable reports. To do so, we would need to use the `keepAlive: true` option when calling eaching of our worklet operations (except perhaps in the last call, if we have no need to extend the worklet's lifetime beyond that call).
+Callers may wish to run multiple worklet operations from the same context, e.g. they might select a URL and then send one or more aggregatable reports. To do so, they would need to use the `keepAlive: true` option when calling each of the worklet operations (except perhaps in the last call, if there was no need to extend the worklet's lifetime beyond that call).
 
-We could make the calls as follows.
-
-In the embedder page:
+As an example, in the embedder page:
 
 ```js
 // Load the worklet module.
--- a/README.md
+++ b/README.md
@@ -113,7 +113,12 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
                 *    The selected URL will be checked to see if it is k-anonymous. If it is not, its k-anonymity will be incremented, but the `default URL` will be returned.
                 *    The reporting metadata will be used in the short-term to allow event-level reporting via `window.fence.reportEvent()` as described in the [FLEDGE explainer](https://github.com/WICG/turtledove/blob/main/Fenced_Frames_Ads_Reporting.md).
             *    There will be a per-origin (the origin of the Shared Storage worklet) budget for `selectURL`. This is to limit the rate of leakage of cross-site data learned from the selectURL to the destination pages that the resulting Fenced Frames navigate to. Each time a Fenced Frame navigates the top frame, for each `selectURL()` involved in the creation of the Fenced Frame, log(|`urls`|) bits will be deducted from the corresponding origin’s budget. At any point in time, the current budget remaining will be calculated as `max_budget - sum(deductions_from_last_24hr)`
-    *   Options can include `data`, an arbitrary serializable object passed to the worklet.
+    *   Options can include:
+        *   `data`, an arbitrary serializable object passed to the worklet. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,7 +113,12 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
             *    There will be a per-origin (the origin of the Shared Storage worklet) budget for `selectURL`. This is to limit the rate of leakage of cross-site data learned from the selectURL to the destination pages that the resulting Fenced Frames navigate to. Each time a Fenced Frame navigates the top frame, for each `selectURL()` involved in the creation of the Fenced Frame, log(|`urls`|) bits will be deducted from the corresponding origin’s budget. At any point in time, the current budget remaining will be calculated as `max_budget - sum(deductions_from_last_24hr)`
-    *   Options can include `data`, an arbitrary serializable object passed to the worklet.
+    *   Options can include:
+        *   `data`, an arbitrary serializable object passed to the worklet. 
+        *   `retain` (defaults to false), a boolean denoting whether the worklet should be retained after it completes work for this call.
+            *   If `retain` is false or not specified, the worklet will shutdown as soon as the operation finishes and subsequent calls to it will fail.
```
jkarlin, line 119: nit: prefer keep-alive as I think that's common parlance and more readily understood.

README.md
```diff
@@ -113,7 +113,12 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
+            *   To continue retaining the worklet throughout multiple calls to `run()` and/or `selectURL()`, each of those calls must include `retain: true` in the `options` dictionary.
```
jkarlin, line 120: s/to continue retaining/to keep the worklet alive/

README.md
```diff
@@ -306,9 +311,68 @@ class ReportingOperation {
 register('send-report', ReportingOperation);
 ```
 
+### Retaining a worklet for multiple operations
+
+We may wish to run multiple worklet operations from the same context, e.g. we might select a URL and then send one or more aggregatable reports. To do so, we would need to use the `retain: true` option when calling eaching of our worklet operations (except perhaps in the last call, if we have no need to extend the worklet's lifetime beyond that call).
```
jkarlin, line 316: editing nit, we seems a bit strange to me. Perhaps, "callers may wish to"? And then below would be something like, s/we could make the calls as follows/As an example:/

README.md
```diff
@@ -306,9 +311,68 @@ class ReportingOperation {
...
+}
+
+register('select-url', URLOperation);
+register('report', ReportOperation);
+```
+
 ## Keep-alive worklet
 
-After a document dies, the corresponding worklet will be kept alive for maximum two seconds to allow the pending operations to execute. This gives more confidence that the end-of-page operations (e.g. reporting) are able to finish.
+After a document dies, the corresponding worklet (if still alive) will continue to be kept alive for a maximum of two seconds to allow the pending operations to execute. This gives more confidence that the end-of-page operations (e.g. reporting) are able to finish.
```
jkarlin, line 375: s/(if still alive)/(if running an operation)/

jkarlin (commented):
Overall looks good, some comments:

README.md
pythagoraskitty: OK, my concern with using "keep-alive" was that it could produce confusion with the "keep-alive" state that happens if the document is destroyed and there are still operations running.  

If you think that this ambiguity is not a problem, then I am happy to change it.

jkarlin: I think this is higher priority in terms of keep-alive term usage, as it's part of the public API. We can rename the "keep-alive worklet" section to something like "Worklets can outlive renderers"

pythagoraskitty: OK, sounds good. Will do, thanks!

README.md

README.md

README.md

pythagoraskitty:
jkarlin@, please take another look.  Thanks!

README.md
```diff
@@ -306,9 +311,66 @@ class ReportingOperation {
...
+class ReportOperation {
+  // See previous examples for how to write a `run()` operation class.
+  async run(data) { ... }
+}
+
+register('select-url', URLOperation);
+register('report', ReportOperation);
+```
+
+## Worklets can outlive renderers
```
xyaoinum, line 371: I think it’s more natural to say that "Worklets can outlive the associated Document", as I’m not sure if "renderer" is a standard term.

xyaoinum (approved):
LGTM % a nit

pythagoraskitty:
I thought that I already changed all of the instances.  Where did I miss?

In articular, I already changed the section heading of "Keep-alive worklet" to "Worklets can outlive the associated document"

jkarlin:
> 

My bad, I missed that change. LGTM

PR #72, opened 2023-04-06
Fix some minor prose errors in spec.bs

pythagoraskitty:


pythagoraskitty:
https://wicg.github.io/shared-storage/pr-preview/refs/pull/72/merge/index.html appears to be correct

PR #75, opened 2023-04-14
Reframe ‘Frequency Capping’ terminology to ‘Creative Selection by Frequency’ 

kevinkiklee:
This PR updates the ‘Frequency Capping/Control’ terminology used in URL selection use cases to minimize confusion and align with industry colloquialism. 

Select URL is a post auction content filtering mechanism that can be used to select one of 8 URLs using cross-site data. One of the key use cases involves storing the number of ads a user has seen across sites in Shared Storage, and using this information to choose the next ad to serve. If the user has seen too many of one ad, they could be shown a new ad or the default ad. 

We were previously referring to this use case as ‘Frequency Capping’ or ‘Frequency Control’. However, we have heard feedback that the industry tends to use those terms when an advertiser is prevented from bidding and prevented from showing the user any ad at all, rather than selecting a different ad. As a result we have reframed the terminology to more explicitly describe the capability, i.e. ‘Creative Selection by Frequency.’


jkarlin:
lgtm!

PR #77, opened 2023-04-21
Only support UTF-8 in response header Byte Sequences

pythagoraskitty:
At least initially, we will only support sending keys and/values via response headers as either [printable-ASCII-character](https://www.rfc-editor.org/rfc/rfc20.html) [Strings](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings), or as [Byte Sequences](https://www.rfc-editor.org/rfc/rfc8941.html#name-byte-sequences) with [UTF-8](https://www.rfc-editor.org/rfc/rfc3629.html) as the intended character set.

The decision to restrict our support of [Byte Sequences](https://www.rfc-editor.org/rfc/rfc8941.html#name-byte-sequences) to a single character encoding will allow us to avoid the additional complexity of either using heuristics to try to determine the intended character encoding, or else requiring that the encoding be specified with the [Byte Sequence](https://www.rfc-editor.org/rfc/rfc8941.html#name-byte-sequence).

We choose to support [UTF-8](https://www.rfc-editor.org/rfc/rfc3629.html) in particular because the [standard preferences it for this purpose](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings).


xyaoinum (approved):
LGTM

PR #94, opened 2023-06-15
Clarify which origin is used for redirects

pythagoraskitty:
We clarify that the origin used for shared storage operations invoked via response headers is the origin of the host server sending the response headers. In particular, in the case of redirects, this is the origin of the redirect URL that is accompanied by the `Shared-Storage-Write` response header(s).


pythagoraskitty:
[xyaoinum](https://github.com/xyaoinum) PTAL, Thanks!

xyaoinum (approved):

PR #103, opened 2023-08-03
Specify IDL attribute `sharedStorageWritable` is boolean

pythagoraskitty:
We specify that the IDL attribute for modifying Shared Storage from response headers is a boolean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
     *   via specifying an content attribute, e.g.: 
         *   `<iframe src=[url] sharedstoragewritable></iframe>`
         *    `<img src=[url] sharedstoragewritable>`
-    *   or via an equivalent IDL attribute, e.g.:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
     *   via specifying an content attribute, e.g.:
```
mikewest, line 207: I was actually talking about the content attribute, not the IDL attribute. Noting that it's a [boolean attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes) means you don't have to specify whether the attribute's value matters.

mikewest (commented):
Thanks for this! One comment inline.

README.md
pythagoraskitty: Ok, just fixed it. (meant to do a PR, but forgot to switch  to my newly created branch and committed the one-word change directly to main, sigh.). In any case, let me know if that's what you had in mind, thanks!

PR #105, opened 2023-08-23
Update spec for enrollment

shivanigithub:
Update spec for enrollment

***
<a href="https://pr-preview.s3.amazonaws.com/shivanigithub/shared-storage/pull/105.html" title="Last updated on Aug 29, 2023, 10:15 PM UTC (6a945b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/105/9d8ae5f...shivanigithub:6a945b9.html" title="Last updated on Aug 29, 2023, 10:15 PM UTC (6a945b9)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -933,7 +933,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
 
     Note: This [=implementation-defined=] condition is intended to allow [=user
         agents=] to disallow the operation for a number of reasons, for example user
-        opt-out or an origin not being
+        opt-out or an origin's site not being
         <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
     1. If |urls| is empty or exceeds the maximum allowed length, return a [=promise rejected=] with a {{TypeError}}.
     1. Let |urlList| be an empty {{list}}.
--- a/spec.bs
+++ b/spec.bs
@@ -376,7 +376,8 @@ The Shared Storage API will integrate into the [=Storage Model|Storage API=] as 
 
     1. Let |origin| be |environment|'s [=url/origin=].
     1. If |origin| is an [=opaque origin=], then return failure.
-    1. If the user has disabled [=shared storage=], then return failure.
+    1. If the user has disabled [=shared storage=] or operation is denied due to other user agent defined mechanism e.g. |origin|'s site not being
+        <a href="https://github.com/privacysandbox/attestation">enrolled</a>, then return failure.
     1. If |shed|[origin] does not exist, then set |shed|[origin] to the result of running [=create a shared storage shelf=] with [=storage type|type=] "`shared`".
     1. Return |shed|[|origin|].
   </div>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -928,6 +929,12 @@ On the other hand, methods for getting data from the [=shared storage database=]
     1. If |worklet|'s [=global scopes|list of global scopes=] is [=list/empty=], then return a [=promise rejected=] with a {{TypeError}}.
     1. [=Assert=] that |worklet|'s [=global scopes|list of global scopes=] [=list/contains=] a single {{SharedStorageWorkletGlobalScope}}.
     1. If the result of running [=SharedStorageWorkletGlobalScope/check whether addModule is finished=] for |worklet|'s {{SharedStorageWorkletGlobalScope}} is false, return a [=promise rejected=]  with a {{TypeError}}.
+    1. Optionally, return a [=promise rejected=] with a {{TypeError}}.
```
shivanigithub, line 932: @xyaoinum @pythagoraskitty I wasn't sure if this is the right place for returning an error from selectUrl and the right error type, can you help confirm if this aligns with the implementation or should this return be at an earlier point like in addModule, thanks!

spec.bs
```diff
@@ -376,7 +376,8 @@ The Shared Storage API will integrate into the [=Storage Model|Storage API=] as
 
     1. Let |origin| be |environment|'s [=url/origin=].
     1. If |origin| is an [=opaque origin=], then return failure.
-    1. If the user has disabled [=shared storage=], then return failure.
+    1. If the user has disabled [=shared storage=] or operation is denied due to other user agent defined mechanism e.g. |origin|'s site not being
```
xyaoinum, line 379: More precisely, I think this should be `|origin|'s [=origin/host=]'s [=host/registrable domain=]`?

spec.bs
xyaoinum: There seems to be two issues:
1. In this spec, we only check the user settings during storage getters/setters, but not for addModule()/selectURL()/run(). We should check the user settings for those as well.
2. In the spec, the Error is always "TypeError". In the code, the error type is sometimes [OperationError](https://webidl.spec.whatwg.org/#operationerror).

I can send a patch to fix (1) to unblock this patch (or @shivanigithub  do you prefer doing the fix along with this patch?). And problem 2 can be addressed separately as it's not only relevant to user setting errors.

spec.bs
xyaoinum: Since you add the link to multiple sites, it seems better to add the link to `<pre class="anchors">` section at the top, and reference it via `[=enrolled=]` in the main text.

```
html; urlPrefix: https://github.com/privacysandbox/attestation
    type: dfn
        text: enrolled
```

Also, I just noticed it's the schemeful site. So, this can be `e.g. the result of [=obtaining a site=] with |origin| not being [=enrolled=]`.

xyaoinum: (FYI I also made a PR to add the settings check for addModule/selectURL/run: https://github.com/WICG/shared-storage/pull/106)

pythagoraskitty:
Couldn't figure out how to reply inline to your comment above, @xyaoinum, i.e.
  
> There seems to be two issues:
> 
> In this spec, we only check the user settings during storage getters/setters, but not for addModule()/selectURL()/run(). We should check the user settings for those as well.
> In the spec, the Error is always "TypeError". In the code, the error type is sometimes [OperationError](https://webidl.spec.whatwg.org/#operationerror).
> I can send a patch to fix (1) to unblock this patch (or @shivanigithub do you prefer doing the fix along with this patch?). And problem 2 can be addressed separately as it's not only relevant to user setting errors.

Anyway, I just approved and merged your fix to #1. 

As for #2, that's due to feedback from my spec mentor, @wanderview, who had recommended that we not delve too deeply into categorizing various exceptions and just use TypeErrors.  But I'm happy to revisit that.

wanderview:
> As for (2.), that's due to feedback from my spec mentor, @wanderview, who had recommended that we not delve too deeply into categorizing various exceptions and just use TypeErrors. But I'm happy to revisit that.

Maybe the solution is to fix the code to throw TypeError to be in sync with the spec?

spec.bs
pythagoraskitty: Edited to add this link.

pythagoraskitty: And used something similar to `the result of [=obtaining a site=] with |origin| not being [=enrolled=]`...

Could probably still be further edited. xyaoinum@ what do you think?

spec.bs
```diff
@@ -236,11 +239,29 @@ Because adding multiple [=module scripts=] via {{Worklet/addModule()}} for the s
 
 When {{Worklet/addModule()}} is called for a worklet, it will run [=check if addModule allowed and update status=], and if the result is false, abort the remaining steps in the {{Worklet/addModule()}} call, as detailed in the [[#worklet-monkey-patch]].
 
+To do so, the {{Worklet}} will need to obtain the result of an [=implementation-defined=] check to [=check if user preference setting allows access to shared storage|determine whether the user preference setting allows access to shared storage=], included as part of an overall check to [=determine whether shared storage is allowed|determine whether shared storage should be allowed=] for a given [=environment=].
```
xyaoinum, line 242: To me, we can drop this intro/transitional paragraph. i.e. the paragraph above calls into [=check if addModule allowed and update status=], and we shouldn't need further clarification; also, we often mention/define algorithms directly without providing an introduction.

xyaoinum (approved):
LGTM % a nit

spec.bs

PR #109, opened 2023-08-31
Update spec.bs to include Permissions Policy

pythagoraskitty:
We define Permissions Policy features for Shared Storage and include checks for whether they are enabled, along with a few other minor tweaks.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/109.html" title="Last updated on Sep 5, 2023, 8:58 PM UTC (cd73167)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/109/0d7bffe...cd73167.html" title="Last updated on Sep 5, 2023, 8:58 PM UTC (cd73167)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -153,9 +153,6 @@ spec: html; urlPrefix: https://en.wikipedia.org/wiki/Entropy_(information_theory
 spec: html; urlPrefix: https://github.com/privacysandbox/attestation
     type: dfn
         text: enrolled
-spec: html; urlPrefix: https://w3c.github.io/webappsec-permissions-policy/
-    type: dfn
-        text: is feature enabled; url: is-feature-enabled
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -941,9 +947,12 @@ On the other hand, methods for getting data from the [=shared storage database=]
     1. Let |resultPromise| be a new [=promise=].
     1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
     1. If |context| is null, return a [=promise rejected=] with a {{TypeError}}.
+    1. Let |document| be |context|'s [=active document=].
+    1. If |document| is not [=fully active=], return a [=promise rejected=] with a {{TypeError}}.
     1. Let |environment| be |context|'s [=active window=]'s [=relevant settings object=].
-    1. If |environment|'s [=associated document=] is not [=fully active=], return a [=promise rejected=] with a {{TypeError}}.
     1. If the result of running [=determine whether shared storage is allowed=] on |environment| is false, return a [=promise rejected=] with a {{TypeError}}.
+    1. Let |origin| be |document|'s [=url/origin=].
+    1. If the result of executing [=is feature enabled|Is feature enabled in document for origin?=] on "[=PermissionsPolicy/shared-storage-select-url=]", |document], and |origin| is "`Disabled`", return false.
```
xyaoinum, line 955: It's simpler to write `If |document| is not [=allowed to use=] the XXX feature, ...`, which already contains the check for "fully active", and uses document's origin by default.

Same for the other place.

xyaoinum (commented):
Looks good % one comment

spec.bs
```diff
@@ -255,7 +252,9 @@ When {{Worklet/addModule()}} is called for a worklet, it will run [=check if add
     1. If |origin| is an [=opaque origin=], then return false.
     1. Let |globalObject| be the [=current realm=]'s [=global object=].
     1. [=Assert=] that |globalObject| is a {{Window}} or a {{SharedStorageWorkletGlobalScope}}.
-    1. If |globalObject| is a {{Window}}, and the result of executing [=is feature enabled|Is feature enabled in document for origin?=] on "[=PermissionsPolicy/shared-storage=]", |globalObject|'s [=associated Document=], and |origin| is "`Disabled`", return false.
+    1. Let |document| be |globalObject|'s [=associated Document=].
```
xyaoinum, line 255: `associated Document` is only a property for `Window`. It seems we should only check the permissions policy when `|globalObject|` is a `Window` (and just skip checking for `SharedStorageWorkletGlobalScope`).

spec.bs

spec.bs

xyaoinum (approved):

PR #110, opened 2023-09-07
Update spec.bs to include operations via HTTP response header

pythagoraskitty:
We update the spec to include how operations can be triggered via HTTP response headers, including by way of `fetch()` as well as HTML images and iframes.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/110.html" title="Last updated on Sep 25, 2023, 6:42 PM UTC (052ea50)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/110/fe6c01d...052ea50.html" title="Last updated on Sep 25, 2023, 6:42 PM UTC (052ea50)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -14,7 +14,6 @@ Abstract: Shared Storage is a storage API that is intentionally not partitioned 
 spec:infra;
     type:dfn;
         text:user agent
-        text:list
         for:/; text:string
 spec:webidl;
     type:interface;
@@ -118,15 +117,16 @@ spec: rfc4648; urlPrefix: https://www.rfc-editor.org/rfc/rfc4648.html
         text: base64 decoding; url: page-5
 spec: rfc8941; urlPrefix: https://httpwg.org/specs/rfc8941.html
     type: dfn
-        text: list structured header; url: list
-        text: parameters; url: param
+        text: structured header; url: specify
+        text: list; for: structured header; url: list
+        text: parameters; for: structured header; url: param
         text: item; for: structured header; url: item
         text: string; for: structured header; url: string
         text: token; for: structured header; url: token
         text: byte sequence; for: structured header; url: binary
         text: boolean; for: structured header; url: boolean
-        text: inner list; url: inner-list
-        text: bare item; url: item
+        text: inner list; for: structured header; url: inner-list
+        text: bare item; for: structured header; url: item
 spec: wikipedia-entropy; urlPrefix: https://en.wikipedia.org/wiki/Entropy_(information_theory)
     type: dfn
         text: bits of entropy
@@ -306,7 +306,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
   The {{SharedStorageRunOperation}} is designed to work with output gates that do not need a return value, like the [=private aggregation=] service. A {{SharedStorageRunOperation}} performs an async operation and returns a promise that resolves to undefined.
 
-  A {{SharedStorageSelectURLOperation}} is an {{SharedStorageOperation}} that takes in a [=list=] of {{SharedStorageUrlWithMetadata}}s (i.e. [=dictionaries=] containing [=strings=] representing [=/URLs=] each wrapped with optional metadata), performs an async operation, and then returns a promise to a {{long}} integer index specifying which of these [=/URLs=] should be selected.
+  A {{SharedStorageSelectURLOperation}} is an {{SharedStorageOperation}} that takes in a [=/list=] of {{SharedStorageUrlWithMetadata}}s (i.e. [=dictionaries=] containing [=strings=] representing [=/URLs=] each wrapped with optional metadata), performs an async operation, and then returns a promise to a {{long}} integer index specifying which of these [=/URLs=] should be selected.
 
   <xmp class='idl'>
     [Exposed=SharedStorageWorklet]
@@ -617,7 +617,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
             1. [=Queue a global task=] on the [=DOM manipulation task source=], given |realm|'s [=global object=], to resolve |promise| with undefined.
             1. Abort these steps.
         1. If |currentValue| is not undefined:
-            1. Let |list| be a new [=list=].
+            1. Let |list| be a new [=/list=].
             1. [=list/Append=] |currentValue| to |list|.
             1. [=list/Append=] |value| to |list|.
             1. Set |value| to the result of running [=string/concatenate=] on |list|.
@@ -724,7 +724,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
 
   To keep track of how this [=navigation entropy allowance=] is used, the [=user agent=] uses a <dfn>shared storage navigation budget table</dfn>, which is a [=map=] of [=calling origins=] to [=navigation entropy ledgers=].
 
-  An <dfn>navigation entropy ledger</dfn> is a [=list=] of [=bit debits=].
+  An <dfn>navigation entropy ledger</dfn> is a [=/list=] of [=bit debits=].
 
   A <dfn>bit debit</dfn> is a [=struct=] containing a {{double}} <dfn for="bit debit">bits</dfn>, indicating a value in [=entropy bits=], along with a {{DOMHighResTimeStamp}} <dfn for="bit debit">timestamp</dfn> (from the [=Unix Epoch=]).
 
@@ -872,7 +872,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
             1. If |operationMap| [=map/contains=] |name|:
                 1.  Let |operation| be |operationMap|[|name|].
                 1. If |options| [=map/contains=] |data|:
-                    1. Let |argumentsList| be a new [=list=].
+                    1. Let |argumentsList| be a new [=/list=].
                     1. [=list/Append=] |data| to |argumentsList|.
                     1. [=Call=] |operation| with |argumentsList|.
                 1. Otherwise, [=call=] |operation| without any arguments list.
@@ -883,7 +883,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
   </div>
 
   <div algorithm>
-    To <dfn>get the select-url result index</dfn>, given {{WindowSharedStorage/worklet}} |worklet|, {{DOMString}} |operationName|, [=list=] |urlList|, and {{SharedStorageRunOperationMethodOptions}} |options|:
+    To <dfn>get the select-url result index</dfn>, given {{WindowSharedStorage/worklet}} |worklet|, {{DOMString}} |operationName|, [=/list=] |urlList|, and {{SharedStorageRunOperationMethodOptions}} |options|:
 
     1. Let |promise| be a new [=promise=].
     1. Let |context| be {{WindowSharedStorage}}'s {{Window}}'s [=Window/browsing context=].
@@ -899,7 +899,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
             1. Let |operationMap| be the associated {{SharedStorageWorkletGlobalScope}}'s [=SharedStorageWorkletGlobalScope/operation map=].
             1. If |operationMap| [=map/contains=] |operationName|:
                 1. Let |operation| be |operationMap|[|operationName|].
-                1. Let |argumentsList| be a new [=list=] with a single entry [=list/contain|containing=] |urlList|.
+                1. Let |argumentsList| be a new [=/list=] with a single entry [=list/contain|containing=] |urlList|.
                 1. If |options| [=map/contains=] |data|, [=list/append=] |data| to |argumentsList|.
                 1. Let |operationResult| be the result of running [=Call=] on |operation| with |argumentsList|.
                 1. If |operationResult| has any error(s), then [=queue a global task=] on the [=DOM manipulation task source=], given |realm|'s [=global object=], to reject |promise| with a {{TypeError}}.
@@ -1048,7 +1048,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
             1. [=Queue a global task=] on the [=DOM manipulation task source=], given |realm|'s [=global object=], to reject |promise| with a {{TypeError}}.
             1. Abort these steps.
         1. If |currentValue| is not undefined:
-            1. Let |list| be a new [=list=].
+            1. Let |list| be a new [=/list=].
             1. [=list/Append=] |currentValue| to |list|.
             1. [=list/Append=] |value| to |list|.
             1. Set |value| to the result of running [=string/concatenate=] on |list|.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1226,6 +1254,282 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+  </div>
+
+
+  <div algorithm>
+    To <dfn>append or modify a Shared-Storage-Writable request header</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. If |request|'s [=request/shared storage writable=] is not true, then abort these steps and return.
+    1. If the result of running [=determine whether a request can use shared storage=] on |request| is false, perform the following steps:
+        1. [=header list/Delete=] "[=Shared-Storage-Writable=]" from |request|'s [=request/header list=].
+        1. Set |request|'s [=request/shared storage writable=] to false.
```
xyaoinum, line 1430: I think we wouldn't want to set `shared storage writable` to false, because it would deny all future redirects. For example, for a redirect chain of A->B->C, C could still be eligible even if B isn't eligible.

spec.bs
```diff
@@ -1226,6 +1254,282 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    1. Return the result of running [=Is feature enabled in document for origin?=] on "[=PermissionsPolicy/shared-storage=]", |document|, and |request|'s [=request/current URL=]'s [=url/origin=].
+
+    Issue: The [=determine whether a request can use shared storage=] algorithm needs to take into account "opt-in features", as articulated in <a href="https://github.com/w3c/webappsec-permissions-policy/pull/499">https://github.com/w3c/webappsec-permissions-policy/pull/499</a>.
+  </div>
+
+
+  <div algorithm>
+    To <dfn>append or modify a Shared-Storage-Writable request header</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. If |request|'s [=request/shared storage writable=] is not true, then abort these steps and return.
```
xyaoinum, line 1427: "abort these steps" should be equivalent to "return", so we only need one of them. "return" seems to be more typical in top level algorithm so is applicable here, per https://github.com/WICG/nav-speculation/pull/202#discussion_r1014553233

xyaoinum (commented):
Looks good % comments. Let's have a spec reviewer to take a look as well.

spec.bs

spec.bs
pythagoraskitty: Interesting. The current implementation would disallow such a chain. Once a request gets marked ineligible due to an ineligible origin in the redirect chain, it cannot currently become eligible again.

See [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/loader/frame_fetch_context.cc;l=360) for `fetch()` and images, and [here](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/navigation_request.cc;l=5111) for iframes.

We could change this in both the implementation and spec, but of course they should match. What do you think?

pythagoraskitty:
wanderview@, PTAL. Thanks!

spec.bs
xyaoinum: Ah I didn't catch this in code review. Yes I think we should change this in both the implementation and spec, so that each redirect is checked independently.

xyaoinum (approved):

pythagoraskitty:
@jyasskin Could you please take a look? I'm trying to get some feedback from a spec expert...  thanks

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    1. For each tuple (|item|, |parameters|) in |operationsToParse|, perform the following steps:
+        1. If |item| is an [=Inner List=], continue.
+        1. [=Assert=] that |item| is an [=Bare Item=].
+        1. Let |operationString| be the result of running [=get the string value=] for |item|.
+        1. If |operationString| is failure, continue.
+        1. Switch on |operationString|:
+            <dl class=switch>
+              <dt> If |operationString| is "`clear`":
+                 <dd> Perform the following steps:
+                     1. Run {{WindowSharedStorage/clear()}} on |sharedStorage|.
```
jyasskin, line 1452: It's generally a bad idea to call WebIDL operations from inside an algorithm, since it leads to ambiguity if JavaScript replaces those functions. ... But it looks like this isn't actually a WebIDL operation even though it's defined `for` a WebIDL interface. In that case, it's probably more clear to just define it as an algorithm that doesn't look like a function call.

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+### [=Shared-Storage-Write=] Response Header ### {#response-header}
+
+  This specification defines a <dfn>Shared-Storage-Write</dfn> HTTP [=/response=] [=header=].
+
+  The [=Shared-Storage-Write=] [=/response=] [=header=] is a [=list structured header=], for which the [=parameters|parameterized=] [=structured header/items=] can have one of the following values, expressed as [=item/token=], [=item/string=], or [=item/byte sequence=]:
+
+    * `set`
+        - required [=parameters=]:
+            * key ([=item/Token=], [=item/String=], or [=item/Byte Sequence=])
+            * value ([=item/Token=], [=item/String=], or [=item/Byte Sequence=])
```
jyasskin, line 1395: This could use a note about why it's important to allow byte sequences in these values. Is it to allow unicode? Also specify that they're interpreted as UTF-8.

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+  When the result of running [=get a structured field value=] for "[=Shared-Storage-Writable=]", "`item`", and [=/request=]'s [=request/header list=] is true, then the [=/response=] to this [=/request=] will be eligible to write to [=shared storage=].
+
+### [=Shared-Storage-Write=] Response Header ### {#response-header}
+
+  This specification defines a <dfn>Shared-Storage-Write</dfn> HTTP [=/response=] [=header=].
+
+  The [=Shared-Storage-Write=] [=/response=] [=header=] is a [=list structured header=], for which the [=parameters|parameterized=] [=structured header/items=] can have one of the following values, expressed as [=item/token=], [=item/string=], or [=item/byte sequence=]:
+
+    * `set`
```
jyasskin, line 1392: Write down what these operations do, in this section. The normative definition is elsewhere, but it's easier to get a sense of things from this section.

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    To <dfn>get the string value</dfn> for a [=Bare Item=] |item|, do the following:
+    1. If the result of running [=check if string-like=] on |item| is false, return failure.
+    1. Let |itemAsString| be the |item|'s underlying [=infra/string=].
+    1. [=Assert=] that |itemAsString| is an [=ASCII string=].
+    1. Switch on the type of |item|:
+        <dl class=switch>
+          <dt> If |item| is a [=item/Token=]:
+            <dd> Return |itemAsString|.
+          <dt> If |item| is a [=item/String=]:
+            <dd> Return the result of [=unwrap a delimited ASCII string|unwrapping a delimited ASCII string=] for |itemAsString| and the double-quote character.
```
jyasskin, line 1497: This would be incorrect because of escaping, but fortunately https://fetch.spec.whatwg.org/#concept-header-list-get-structured-header already converted these to strings.

jyasskin (changes requested):

spec.bs
```diff
@@ -42,17 +42,21 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
-spec: html; urlPrefix: https://dom.spec.whatwg.org/
+spec: dom; urlPrefix: https://dom.spec.whatwg.org/
     type: dfn
         text: origin; for: document; url: concept-document-origin
-spec: html; urlPrefix: https://infra.spec.whatwg.org
+spec: infra; urlPrefix: https://infra.spec.whatwg.org
     type: dfn
         text: user agent; url: user-agent
-        text: string; url: string
+        text: string; for: infra; url: string
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -137,22 +141,46 @@ spec: html; urlPrefix: https://wicg.github.io/fenced-frame/
...
-spec: html; urlPrefix: https://github.com/privacysandbox/attestation
+spec: attestation; urlPrefix: https://github.com/privacysandbox/attestation
     type: dfn
         text: enrolled
+spec: rfc8941; urlPrefix: https://httpwg.org/specs/rfc8941.html
+    type: dfn
+        text: list structured header; url: list
+        text: parameters; url: param
+        text: item; for: structured header; url: item
+        text: string; for: item; url: string
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    add the step
+
+    1. [=Handle a Shared-Storage-Write response=], given [=/response=] |internalResponse| and [=/request=] <var ignore=''>request</var> as input.
+  </div>
+
+## Shared Storage HTTP Headers ## {#headers}
+
+### [=Shared-Storage-Writable=] Request Header ### {#request-header}
+
+  This specification defines a <dfn>Shared-Storage-Writable</dfn> HTTP [=/request=] [=header=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+  The [=Shared-Storage-Writable=] [=/request=] [=header=] is a [=item/boolean|boolean structured header=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+<span class=todo>Add examples.</span>
+
+## Shared Storage Fetch-Related Algorithms ## {#ss-fetch-algo}
+
+  <div algorithm>
+    To <dfn>determine whether a request can currently use shared storage</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. Set |window| to |request|’s [=request/window=].
+    1. If |window| is not an [=environment settings object=] whose [=global object=] is a {{Window}}, return false.
```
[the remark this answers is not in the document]
pythagoraskitty: Totally agree. i found this strange to say the least!

spec.bs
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    1. Return the result of running [=Is feature enabled in document for origin?=] on "[=PermissionsPolicy/shared-storage=]", |document|, and |request|'s [=request/current URL=]'s [=url/origin=].
+
+    Issue: The [=determine whether a request can currently use shared storage=] algorithm needs to take into account "opt-in features", as articulated in <a href="https://github.com/w3c/webappsec-permissions-policy/pull/499">https://github.com/w3c/webappsec-permissions-policy/pull/499</a>.
+  </div>
+
+
+  <div algorithm>
+    To <dfn>append or modify a Shared-Storage-Writable request header</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. If |request|'s [=request/shared storage writable=] is not true, then return.
```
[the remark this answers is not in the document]
pythagoraskitty: Added a note.

spec.bs
pythagoraskitty: Replaced

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+            <dd> Return the result of [=unwrap a delimited ASCII string|unwrapping a delimited ASCII string=] for |itemAsString| and the double-quote character.
+          <dt> If |item| is a [=item/Byte Sequence=]:
+            <dd> Perform the following steps:
+                1. Let |innerItem| be the result of [=unwrap a delimited ASCII string|unwrapping a delimited ASCII string=] for |itemAsString| and "`:`".
+                1. If |innerItem| is failure, return failure.
+                1. Let |bytes| be the [=base64 decoding=] of |innerItem|.
+                1. If |bytes| is an error, return failure.
+                1. Let |bytesFromUTF8| be the result of running [=UTF-8 decode=] on |bytes|.
+                1. If |bytesFromUTF8| is an error, return failure.
+                1. Return the result of [=encoding UTF-16=] on |bytesFromUTF8|.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    * `clear`
+
+<span class=todo>Add examples.</span>
+
+## Shared Storage Fetch-Related Algorithms ## {#ss-fetch-algo}
+
+  <div algorithm>
+    To <dfn>determine whether a request can currently use shared storage</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. Set |window| to |request|’s [=request/window=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+### [=Shared-Storage-Write=] Response Header ### {#response-header}
+
+  This specification defines a <dfn>Shared-Storage-Write</dfn> HTTP [=/response=] [=header=].
+
+  The [=Shared-Storage-Write=] [=/response=] [=header=] is a [=list structured header=], for which the [=parameters|parameterized=] [=structured header/items=] can have one of the following values, expressed as [=item/token=], [=item/string=], or [=item/byte sequence=]:
+
+    * `set`
+        - required [=parameters=]:
+            * key ([=item/Token=], [=item/String=], or [=item/Byte Sequence=])
```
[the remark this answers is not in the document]
pythagoraskitty: Yes, we want to be able to have Unicode keys/values. Added a note, as suggested in your other comment.

spec.bs
pythagoraskitty: Added

spec.bs

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+        <dl class=switch>
+          <dt> If |item| is a [=item/Token=]:
+            <dd> Return |itemAsString|.
+          <dt> If |item| is a [=item/String=]:
+            <dd> Return the result of [=unwrap a delimited ASCII string|unwrapping a delimited ASCII string=] for |itemAsString| and the double-quote character.
+          <dt> If |item| is a [=item/Byte Sequence=]:
+            <dd> Perform the following steps:
+                1. Let |innerItem| be the result of [=unwrap a delimited ASCII string|unwrapping a delimited ASCII string=] for |itemAsString| and "`:`".
+                1. If |innerItem| is failure, return failure.
+                1. Let |bytes| be the [=base64 decoding=] of |innerItem|.
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+  The [=Shared-Storage-Writable=] [=/request=] [=header=] is a [=item/boolean|boolean structured header=].
+
+  When the result of running [=get a structured field value=] for "[=Shared-Storage-Writable=]", "`item`", and [=/request=]'s [=request/header list=] is true, then the [=/response=] to this [=/request=] will be eligible to write to [=shared storage=].
+
+### [=Shared-Storage-Write=] Response Header ### {#response-header}
+
+  This specification defines a <dfn>Shared-Storage-Write</dfn> HTTP [=/response=] [=header=].
+
+  The [=Shared-Storage-Write=] [=/response=] [=header=] is a [=list structured header=], for which the [=parameters|parameterized=] [=structured header/items=] can have one of the following values, expressed as [=item/token=], [=item/string=], or [=item/byte sequence=]:
```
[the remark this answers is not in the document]
pythagoraskitty: Yes, we want to be able to allow non-ASCII Unicode.

spec.bs
```diff
@@ -1226,6 +1221,273 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    * `append`
+        - required [=structured header/parameters=]:
+            * key ([=structured header/Token=], [=structured header/String=], or [=structured header/Byte Sequence=])
+            * value ([=structured header/Token=], [=structured header/String=], or [=structured header/Byte Sequence=])
+    * `delete`
+        - required [=structured header/parameters|parameter=]:
+            * key ([=structured header/Token=], [=structured header/String=], or [=structured header/Byte Sequence=])
+    * `clear`
+
+    Note: We allow [=structured header/Byte Sequences=] in order to accommodate Unicode keys and values. Any [=structured header/Byte Sequence=] will be assumed to be [=UTF-8 encoded=] and will fail to parse otherwise.
```
jyasskin, line 1374: Consider using https://httpwg.org/http-extensions/draft-ietf-httpbis-sfbis.html#name-display-strings (new in the latest draft of Structured Fields, and not in an RFC yet) instead of byte sequences to store Unicode. Note the disagreements about adding them in https://github.com/httpwg/http-extensions/pull/2494 and https://lists.w3.org/Archives/Public/ietf-http-wg/2023AprJun/thread.html#msg177. It might be worth mailing the HTTPWG with this use case to get their opinions on how to encode what you want to encode.

pythagoraskitty: Thanks for bring to my attention. 

Oh wow, display strings seem very unsafe, and I agree with those asking why they are needed given that we already have the safer byte sequences...

I can discuss with my team, but I would definitely want to have an in-depth conversation with some Chromium network security people before switching over to display strings, if we do that. I can put this on our todo list for potential changes that need research first, in case display strings are adopted...

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+    Issue: The [=determine whether a request can currently use shared storage=] algorithm needs to take into account "opt-in features", as articulated in <a href="https://github.com/w3c/webappsec-permissions-policy/pull/499">https://github.com/w3c/webappsec-permissions-policy/pull/499</a>.
+  </div>
+
+
+  <div algorithm>
+    To <dfn>append or modify a Shared-Storage-Writable request header</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. If |request|'s [=request/shared storage writable=] is not true, then return.
+    1. If the result of running [=determine whether a request can currently use shared storage=] on |request| is false, [=header list/delete=] "[=Shared-Storage-Writable=]" from |request|'s [=request/header list=].
+    1. Otherwise, [=header list/set=] ("[=Shared-Storage-Writable=]", "?1") in |request|'s [=request/header list=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+## Shared Storage HTTP Headers ## {#headers}
+
+### [=Shared-Storage-Writable=] Request Header ### {#request-header}
+
+  This specification defines a <dfn>Shared-Storage-Writable</dfn> HTTP [=/request=] [=header=].
+
+  The [=Shared-Storage-Writable=] [=/request=] [=header=] is a [=item/boolean|boolean structured header=].
+
+  When the result of running [=get a structured field value=] for "[=Shared-Storage-Writable=]", "`item`", and [=/request=]'s [=request/header list=] is true, then the [=/response=] to this [=/request=] will be eligible to write to [=shared storage=].
```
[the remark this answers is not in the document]

spec.bs
```diff
@@ -1226,6 +1254,280 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+
+    1. If |request|'s [=request/shared storage writable=] is not true, then return.
+    1. If the result of running [=determine whether a request can currently use shared storage=] on |request| is false, [=header list/delete=] "[=Shared-Storage-Writable=]" from |request|'s [=request/header list=].
+    1. Otherwise, [=header list/set=] ("[=Shared-Storage-Writable=]", "?1") in |request|'s [=request/header list=].
+  </div>
+
+  <div algorithm>
+    To <dfn>handle a Shared-Storage-Write response</dfn>, given a [=/response=] |response| and a [=/request=] |request|, perform the following steps:
+
+    1. If |request|'s [=request/header list=] does not [=list/contain=] "[=Shared-Storage-Writable=]", then return.
```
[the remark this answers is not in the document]

spec.bs
pythagoraskitty: Done.

spec.bs
pythagoraskitty: Added issue #117 .

pythagoraskitty:
Landing so that I can open another PR on the spec. Will continue to iterate.

spec.bs
```diff
@@ -1226,6 +1222,282 @@ On the other hand, methods for getting data from the [=shared storage database=]
...
+## Shared Storage Fetch-Related Algorithms ## {#ss-fetch-algo}
+
+  <div algorithm>
+    To <dfn>determine whether a request can currently use shared storage</dfn>, given a [=/request=] |request|, perform the following steps:
+
+    1. Let |window| to |request|’s [=request/window=].
+    1. If |window| is not an [=environment settings object=] whose [=global object=] is a {{Window}}, return false.
+    1. If the result of running [=determine whether shared storage is allowed=] for |window| is false, return false.
+    1. Let |document| be |window|'s [=global object=]'s [=associated document=].
+    1. Return the result of running [=Is feature enabled in document for origin?=] on "[=PermissionsPolicy/shared-storage=]", |document|, and |request|'s [=request/current URL=]'s [=url/origin=].
```
jyasskin, line 1396: This should probably use [[=Should request be allowed to use feature=]](https://www.w3.org/TR/permissions-policy/#should-request-be-allowed-to-use-feature)? instead of "Is feature enabled...".

PR #118, opened 2023-09-25
Update spec.bs to Change From Per-Origin to Per-Site Budgets

pythagoraskitty:
We update the spec to reflect the explainer changes landed in PR #112 .

We are now going to use per-site privacy budgets instead of per-origin privacy budgets. This will better align with enrollment enforcement, which is also per-site.


xyaoinum (approved):

spec.bs
```diff
@@ -804,10 +804,11 @@ On the other hand, methods for getting data from the [=shared storage database=]
     1. Let |node| be |sourceDocument|'s [=node navigable=].
     1. While |node| is not null:
         1. If |node| has a "fenced frame config struct":
-            1. Let |origin| be |node|'s [=active document=]'s [=document/origin=].
+            1. Let |site| be the result of running [=obtain a site=] with |node|'s [=active document=]'s [=document/origin=].
+            1. [=Assert=] that |site| is not an [=opaque origin=].
```
jyasskin, line 808: Why can't |node| have an opaque origin?

jyasskin (approved):

spec.bs
pythagoraskitty: Good point. I removed that assumption, and put in a condition to skip the case where it's opaque.

PR #120, opened 2023-10-20
Update explainer: Add 'Sec-' prefix to request header

pythagoraskitty:
The request header added to requests that are eligible to write to shared storage via response header ought to be a "forbidden header" according to the [Fetch specification](https://fetch.spec.whatwg.org/#forbidden-request-header), i.e. a header that JS cannot add or modify.

We were using `Shared-Storage-Writable` as the name of the request header, which was not a forbidden header and could be modified by JS.

This commit updates the explainer to add the prefix `Sec-` in order to make this header forbidden. The new request header attached to eligible outgoing requests will be `Sec-Shared-Storage-Writable: ?1`.

We will update the specification draft accordingly in #121.


pythagoraskitty:
@xyaoinum PTAL, thanks!

xyaoinum (approved):

PR #128, opened 2023-12-18
Make Shared-Storage-Write a forbidden response header

pythagoraskitty:
We update spec.bs to include in the Fetch monkey patch that the new '`Shared-Storage-Write`' header should be added to the list of forbidden response headers. 

This will ensure that the '`Shared-Storage-Write`' response header can only be set by the server and read by the user agent. Frontend JavaScript will not be able to read, write, or modify it.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/128.html" title="Last updated on Dec 18, 2023, 6:58 PM UTC (98b7ad6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/128/d19acf9...98b7ad6.html" title="Last updated on Dec 18, 2023, 6:58 PM UTC (98b7ad6)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1346,7 +1346,7 @@ The IDL attribute {{HTMLSharedStorageWritableElementUtils/sharedStorageWritable}
 
 ### [:Sec-Shared-Storage-Writable:] Request Header ### {#request-header}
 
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
Decided to not do this.

PR #142, opened 2024-03-27
[explainer update] Add "Shared-Storage-Worklet-Allowed: ?1" response header requirement

xyaoinum:
Add the "Shared-Storage-Worklet-Allowed: ?1" response header requirement. This provides an additional layer of security compared to relying solely on CORS.


jkarlin:
Issue #2

PR #156, opened 2024-05-29
[spec] Allow createWorklet() in opaque origin contexts

xyaoinum:
In opaque origin contexts (e.g. sandboxed iframes), we now allow `window.createWorklet(url)` and allow `selectURL()`/`run()` on the resulting worklet object. Those methods were previously disallowed in opaque origin contexts.

Reason: While direct storage access is typically disallowed in sandboxed iframe, that's more of a side effect of having an opaque origin, rather than the original intent of sandboxing. Thus, it should be okay to allow creating new non-opaque contexts via `createWorklet(url)`, and the new contexts can further access shared storage. This enables more flexible functionality within the sandboxed environment while still adhering to the security principles of sandboxing.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/156.html" title="Last updated on Jun 6, 2024, 10:30 PM UTC (877a19c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/156/a8fcc13...877a19c.html" title="Last updated on Jun 6, 2024, 10:30 PM UTC (877a19c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -276,7 +276,7 @@ When {{Worklet/addModule()}} is called for a worklet, it will run [=check if add
     - For creating a worklet, |environment| is the [=environment settings object=] associated with the {{Window}} that created the worklet, and |origin| is the module script url's [=url/origin=].
     - For running operations on a worklet (from a {{Window}}), and for each method under [[#worklet-setter]] (from {{SharedStorageWorkletGlobalScope}}), |environment| is the [=environment settings object=] associated with the {{Window}} that created the worklet, and |origin| is the worklet's [=global scopes=][0]'s [=global object/realm=]'s [=realm/settings object=]'s [=environment settings object/origin=].
     - For [[#ss-fetch-algo]], |environment| is the request's [=request/window=], and |origin| is the request's [=request/current URL=]'s [=url/origin=].
-    - For [[#ss-fetch-algo]], for {{WindowSharedStorage/createWorklet()}}, and for {{SharedStorageWorklet/selectURL()}} and {{SharedStorageWorklet/run()}} that operate on a worklet created from {{WindowSharedStorage/createWorklet()}}, |allowedInOpaqueOriginContext| is true. For other methods, |allowedInOpaqueOriginContext| is false.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -277,16 +276,19 @@ When {{Worklet/addModule()}} is called for a worklet, it will run [=check if add
...
     - For running operations on a worklet (from a {{Window}}), and for each method under [[#worklet-setter]] (from {{SharedStorageWorkletGlobalScope}}), |environment| is the [=environment settings object=] associated with the {{Window}} that created the worklet, and |origin| is the worklet's [=global scopes=][0]'s [=global object/realm=]'s [=realm/settings object=]'s [=environment settings object/origin=].
     - For [[#ss-fetch-algo]], |environment| is the request's [=request/window=], and |origin| is the request's [=request/current URL=]'s [=url/origin=].
-        - Note that in this scenario, a slightly modified algorithm [=determine whether using shared storage in a request is allowed by context=] is used in place of [=determine whether shared storage is allowed by context=].
+    - For [[#ss-fetch-algo]], for {{WindowSharedStorage/createWorklet()}}, and for {{SharedStorageWorklet/selectURL()}} and {{SharedStorageWorklet/run()}} that operate on a worklet created from {{WindowSharedStorage/createWorklet()}}, |allowedInOpaqueOriginContext| is true. For other methods, |allowedInOpaqueOriginContext| is false.
```
pythagoraskitty, line 279: Technically, only for createWorklet() called for a cross-origin worklet script, and run()/selectURL() that operate on a cross-origin worklet created from createWorklet...

pythagoraskitty: Suggest something like:
```
For [[#ss-fetch-algo]], for {{WindowSharedStorage/createWorklet()}} called on a cross-origin worklet script, and for {{SharedStorageWorklet/selectURL()}} and {{SharedStorageWorklet/run()}} that operate on a cross-origin worklet created from {{WindowSharedStorage/createWorklet()}}, |allowedInOpaqueOriginContext| is true. For other methods, |allowedInOpaqueOriginContext| is false.
```

xyaoinum: Done.

Issue #127, opened 2023-11-30
make cross-site information accessible


fcurti:
I need to share informations on several domains.

I have to accomplish, for example, the following use case

A login web page (a.login) write a cookie with some informations.
Some information needs to be retrieved on other properties ( a.myproperties, b.myproperties) where user navigates.

In login web page and other properties I can inject javascript (obviously).

I'm asking which is the best strategy to achieve this using the shared storage API.

pythagoraskitty:
When you say:
> Some information needs to be retrieved on other properties ( a.myproperties, b.myproperties) where user navigates.

Can you be a little more specific about what kind of information retrieval you will need?

Keep in mind that it's easy to write to shared storage, but there are intentional restrictions on reading from shared storage. Currently, there are two output gates: private aggregation reports and URL selection.

If I get a better sense of what kind of information you are planning to retrieve and/or how you are hoping to use it, I might be able to provide better suggestions of how to accomplish your task.

fcurti:
@pythagoraskitty 
for example we need to retrieve the user PPID.

During user login on a.login the login service creates a cookie with ppid.
We need to read this ppid on other properties, for example a.myproperties1, b.myproperties2 and so on.

In this time we are using a 3rd party request to save the ppid as a cookie on a single remote domain (shared.domain)
As well we are using a 3rd party request on shared.domain to read the ppid and save the ppid in a cookie on domains where user navigates.

b-hatley:
Hi @pythagoraskitty, our company (Flashtalking) is also seeing the same issue, causing a major bump in our road to integrating with PS APIs. Below is an overview of our challenge:

As a creative ad server, we need to serve creatives from a CDN (cdn.flashtalking.com). At the same time, we also need the ability to read data in Shared Storage we previously set on the browser to personalize the creative during the ad request. 

Currently, we are unable to complete this because our ad server engine, which is responsible for writing to Shared Storage and decisioning on the data within, runs on a secondary sub-domain ([servedby.flashtalking.com](http://servedby.flashtalking.com/)).

Example flow:
* User 123 visit Advertiser A’s site and looks at shoes.
* Flashtalking sets "shoes" segment in Shared Storage with [servedby.flashtalking.com](http://servedby.flashtalking.com/)  
* Flashtalking receives an ad request from Publisher XYZ for Advertiser A from User 123’s browser
[...]

PR #158, opened 2024-06-12
Explainer: Allow cross-origin script in addModule & align createWorklet

pythagoraskitty:
Currently, `addModule` only allows same-origin script. This was for convenience of the initial implementation, however, and is no longer necessary. 

The [worklet standard](https://html.spec.whatwg.org/multipage/worklets.html#dom-worklet-addmodule) does not contain this restriction. In fact, we have received [feedback](https://github.com/WICG/shared-storage/issues/127) from developers stating they would like to be able to host and run their worklet script on a separate origin---say a CDN---from the origin that owns and writes their shared storage data.

So we update the explainer to remove the same-origin restriction for `addModule`. We also note that, when the worklet script is cross-origin to the invoking context, the invoking context's origin is used as the partition origin for accessing shared storage.

Since `createWorklet` already allows cross-origin scripts, but currently uses the worklet script's origin as the data partition origin, updating `addModule` as described above without also making a change to `createWorklet` is liable to cause developer confusion in the long term.

We have therefore decided to align `createWorklet`'s default data partition origin with `addModule`'s. `createWorklet` will use the invoking context's origin by default. This is a breaking change, but current usage of `createWorklet` is low as it was just introduced in M125.

To preserve the ability to create a worklet whose script is cross-origin to the invoking context and then run operations on the worklet script origin's shared storage data, we introduce a new `dataOrigin` option for `createWorklet`. Passing `dataOrigin` with "script-origin" as its value will designate the worklet script origin as the partition origin for shared storage. For now, "script-origin" and "context-origin" will be the only allowed values for `dataOrigin`, when used. We’re considering adding support for separate data and script origins for `createWorklet` in the future.

A corresponding specification update will follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,6 +463,8 @@ register('report', ReportOperation);
 
 There are currently four (4) approaches to creating a worklet that loads cross-origin script. The partition origin for the worklet's shared storage data access depends on the approach.
 
+#### Using the context origin as data partition origin
+The first three (3) approaches use the invoking context's origin as the partition origin for shared storage data access and the invoking context's site for shared storage budget withdrawals.
 
 1. Call `addModule()` with a cross-origin script. 
 
@@ -493,6 +495,9 @@ There are currently four (4) approaches to creating a worklet that loads cross-o
     ```
 
     For any subsequent `run()` or `selectURL()` operation invoked on this worklet, the shared storage data for "https://a.example" (i.e. the context origin) will be used.
+
+#### Using the worklet script origin as data partition origin
+The fourth approach uses the worklet script's origin as the partition origin for shared storage data access and the worklet script's site for shared storage budget withdrawals.
 
 4. Call `createWorklet()` with a cross-origin script, setting its `dataOption` to the worklet script's origin.
 
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
         *   Currently, the `dataOrigin` option, if used, is restricted to having either "script-origin" or "context-origin" as its value. "script-origin" designates the worklet script origin as the data partition origin; "context-origin" designates the invoking context origin as the data partition origin.
     *   The object that the returned Promise resolves to has the same type with the implicitly constructed `window.sharedStorage.worklet`. However, for a worklet created via `window.sharedStorage.createWorklet(url, options)`, only `selectURL()` and `run()` are available, whereas calling `addModule()` will throw an error. This is to prevent leaking shared storage data via `addModule()`, similar to the reason why `addModule()` can only be invoked once on the implicitly constructed `window.sharedStorage.worklet`.
     *   Redirects are not allowed.
-    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin and the `dataOrigin` option is different from the context origin, a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
+    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin and when `dataOrigin` is "script-origin", a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
     *   The script server must carefully consider the security risks of allowing worklet creation by other origins (via `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` and CORS), because this will also allow the worklet creator to run subsequent operations, and a malicious actor could poison and use up the worklet origin's budget.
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -455,6 +459,41 @@ register('select-url', URLOperation);
 register('report', ReportOperation);
 ```
 
+### Using cross-origin worklets
```
jkarlin, line 462: This section is a bit confusing, as it's using cross-origin to mean different things. Use cases 1 and 2 are about cross-origin worklet script. Use case 3 is about a cross-origin data origin.

Can you instead make 2 sections, one titled "Loading cross-origin worklet scripts" and another, "Creating worklets with cross-origin data origins" or some such?

pythagoraskitty: I changed the section title to "Loading cross-origin worklet scripts" and slightly revised the content. 

I think that it would make more sense to wait to add a separate "Creating worklets with cross-origin data origins" section until we decide to allow other values for the `dataOrigin` option (if we do).

Anyway, please take a look and see if the revised section looks good. If you would still like it in 2 sections, I can split it then. Thanks.

jkarlin: | I think that it would make more sense to wait to add a separate "Creating worklets with cross-origin data origins" section until we decide to allow other values for the dataOrigin option (if we do).

Using `script-origin` _is_ a cross-origin data origin from the calling context.

dmdabbs:
If a publisher allows some partner to use SharedStorage via the policy-controlled feature, is there a way to restrict the partner's data access?

dmdabbs:
Will the new `dataOrigin` capability be feature-detectable?

GorAbgaryan1611 (approved):

README.md
```diff
@@ -147,10 +149,12 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
     *   The object that the returned Promise resolves to has the same type with the implicitly constructed `window.sharedStorage.worklet`. However, for a worklet created via `window.sharedStorage.createWorklet(url, options)`, only `selectURL()` and `run()` are available, whereas calling `addModule()` will throw an error. This is to prevent leaking shared storage data via `addModule()`, similar to the reason why `addModule()` can only be invoked once on the implicitly constructed `window.sharedStorage.worklet`.
     *   Redirects are not allowed.
-    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin, a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
+    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin and the `dataOrigin` option is different from the context origin, a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
```
xyaoinum, line 157: Would it be clearer to say: and <code>when \`dataOrigin\` is "script-origin"</code>?

README.md
```diff
@@ -455,6 +459,56 @@ register('select-url', URLOperation);
...
+
+#### Using the context origin as data partition origin
+The first three (3) approaches use the invoking context's origin as the partition origin for shared storage data access and the invoking context's site for shared storage budget withdrawals.
+
+1. Call `addModule()` with a cross-origin script. 
+
+    In an "https://a.example" context in the embedder page:
+
+    ```
+    await sharedStorage.addModule("https://b.example/worklet.js");
```
xyaoinum, line 474: `sharedStorage.worklet.addModule`

README.md

README.md

README.md
pythagoraskitty: So does the revised section look good for the current proposal? Or are there any changes that we should make in this PR?

pythagoraskitty:
> If a publisher allows some partner to use SharedStorage via the policy-controlled feature, is there a way to restrict the partner's data access?

The publisher/embedder can control which origins have access to Shared Storage via the [Permissions Policy](https://www.w3.org/TR/permissions-policy/), yes. The origins allowed to use the [policy-controlled feature "shared-storage"](https://github.com/WICG/shared-storage/blob/cammie-branch3/README.md#permissions-policy) are those that will have permission to be used as data origins for the Shared Storage API.


> Will the new `dataOrigin` capability be feature-detectable?

You will be able to detect it if you call `createWorklet()` with a valid same-origin worklet script as the source but an invalid value passed to "dataOrigin", e.g. something like

```
let dataOriginOptionDefined = false;
try {
  await sharedStorage.createWorklet("dummy-worklet.js", {"dataOrigin": "invalid"});
} catch (e) {
  if (e.message.includes("dataOrigin")) {  // or some other check to help distinguish it from other possible errors
    dataOriginOptionDefined = true;
  }
}
```

We haven't implemented this yet, and the exactly error message/type haven't been determined yet. So the above snippet is subject to change (and I didn't test it), but something along these lines should work.

xyaoinum (approved):

PR #131, opened 2024-01-18
Spec update: Support multiple & cross-origin worklets

xyaoinum:
- selectURL() and run() will be exposed to the shared storage worklet object. When calling on the default scoped worklet (i.e. sharedStorage.worklet.selectURL()/run()), the behavior is equivalent to sharedStorage.selectURL()/run().
- Users can create new worklets via `const worklet = await sharedStorage.createWorklet(url, options)`. This can be used to start multiple and potentially cross-origin worklets from a single document.
- User settings error won't be exposed to the caller (i.e. will be treated as success) in the case of creating or using a cross-origin worklet. This is to prevent leaking user's settings for an arbitrary site.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/131.html" title="Last updated on Apr 13, 2024, 2:47 AM UTC (9100129)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/131/ea33aba...9100129.html" title="Last updated on Apr 13, 2024, 2:47 AM UTC (9100129)">Diff</a>

PR #161, opened 2024-06-18
Spec: Allow cross-origin script in addModule & align createWorklet

pythagoraskitty:
We make the spec changes to accompany #158 . 

In particular, we make it possible for `sharedStorage.worklet.addModule()` to create a worklet with cross-origin worklet script. We also align the behavior of `sharedStorage.createWorklet()` with `addmodule()` so that it uses the invoking context's origin as the data partition origin by default, while adding a mechanism to use the script's origin instead if specified.

For a more detailed description, please see #158 .

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/161.html" title="Last updated on Jul 17, 2024, 6:16 PM UTC (420fcc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/161/919562d...420fcc5.html" title="Last updated on Jul 17, 2024, 6:16 PM UTC (420fcc5)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -478,7 +478,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
   ### Monkey Patch for [=HTTP fetch=] ### {#http-fetch-monkey-patch}
   The following step will be added to the [=HTTP fetch=] steps, before checking the redirect status (i.e. "6. If |internalResponse|'s status is a redirect status, ..."):
 
-  1. If |request|'s [=request/destination=] is "sharedstorageworklet":
+  1. If |request|'s [=request/destination=] is `"xoriginsharedstorageworklet"`:
       1. [=Assert=]: |request|'s [=request/origin=] is not "<code>client</code>".
       1. If |request|'s [=request/origin=] and |request|'s [=request/URL=]'s [=url/origin=] are not [=same origin=]:
           1. Let |headers| be |internalResponse|'s [=response/header list=].
@@ -520,11 +520,11 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
   The {{SharedStorageWorklet}}'s [=worklet global scope type=] is {{SharedStorageWorkletGlobalScope}}.
 
-  The {{SharedStorageWorklet}}'s [=worklet destination type=] is either `"sharedstorageworklet"` or `"sharedstorageworkletcontextorigin"`. It is set to `"sharedstorageworkletcontextorigin"` by default.
+  The {{SharedStorageWorklet}}'s [=worklet destination type=] is either `"sharedstorageworklet"` or `"xoriginsharedstorageworklet"`. It is set to `"sharedstorageworklet"` by default.
 
   ### Monkey Patch for request [=request/destination=] ### {#request-destination-monkey-patch}
 
-  The fetch request's [=request/destination=] field should additionally include `"sharedstorageworklet"` and `"sharedstorageworkletcontextorigin"` as valid values.
+  The fetch request's [=request/destination=] field should additionally include `"sharedstorageworklet"` and `"xoriginsharedstorageworklet"` as valid values.
 
   <xmp class='idl'>
     callback RunFunctionForSharedStorageSelectURLOperation = Promise<unsigned long>(sequence<USVString> urls, optional any data);
@@ -1150,7 +1150,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
 
     1. Let |sharedStorageWorklet| be a new {{SharedStorageWorklet}}.
     1. If |options| [=map/contains=] |dataOrigin|, set |sharedStorageWorklet|'s [=SharedStorageWorklet/data origin=] to |options|[|dataOrigin|].
-    1. If |sharedStorageWorklet|'s [=SharedStorageWorklet/data origin=] is not `"context-origin"`, set |sharedStorageWorklet|'s [=worklet destination type=] to `"sharedstorageworklet"`.
+    1. If |sharedStorageWorklet|'s [=SharedStorageWorklet/data origin=] is not `"context-origin"`, set |sharedStorageWorklet|'s [=worklet destination type=] to `"xoriginsharedstorageworklet"`.
     1. Let |addModulePromise| be the result of invoking sharedStorageWorklet.{{Worklet/addModule()|addModule}}(|moduleURL|, |options|).
     1. Let |resultPromise| be a new [=promise=].
     1. [=Upon fulfillment=] of |addModulePromise|, [=resolve=] |resultPromise| to |sharedStorageWorklet|.
--- a/spec.bs
+++ b/spec.bs
@@ -485,7 +485,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
           1. Let |allowed| be the result of running [=get a structured field value=] algorithm given \`<a http-header><code>Shared-Storage-Cross-Origin-Worklet-Allowed</code></a>\`, "item", and |headers| as input.
           1. If |allowed| is false, then return a [=network error=].
 
-  Note: It is the responsibility of the site serving the module script to carefully consider the security implications: when the module script's [=/URL=]'s [=url/origin=] and the worklet's creator {{Window}} origin are not [=same origin=], by sending permissive CORS headers and the \`<a http-header><code>Shared-Storage-Cross-Origin-Worklet-Allowed</code></a>\` header on the module script response, the server will be granting the worklet's creation and subsequent operations on the worklet. For example, the worklet's creator {{Window}} could poison and use up the worklet origin's [=remaining navigation budget=] by calling {{SharedStorageWorklet/selectURL()}} or {{SharedStorageWorklet/run()}}, where the worklet origin is the global scope's [=global object/realm=]'s [=realm/settings object=]'s [=environment settings object/origin=].
+  Note: It is the responsibility of the site serving the module script to carefully consider the security implications: when the module script's [=/URL=]'s [=url/origin=] and the worklet's creator {{Window}} origin are not [=same origin=], by sending permissive CORS headers the \`<a http-header><code>Shared-Storage-Cross-Origin-Worklet-Allowed</code></a>\` header on the module script response, the server will be granting the worklet's creation and subsequent operations on the worklet, while allowing the worklet to use the worklet's script's [=url/origin=] as the [=url/origin=] for accessing the shared storage data, i.e. the [=url/origin=] used in all calls to [=obtain a shared storage bottle map=]. For example, the worklet's creator {{Window}} could poison and use up the worklet origin's [=remaining navigation budget=] by calling {{SharedStorageWorklet/selectURL()}} or {{SharedStorageWorklet/run()}}, where the worklet origin is the global scope's [=global object/realm=]'s [=realm/settings object=]'s [=environment settings object/origin=].
 
   ### Monkey Patch for {{Worklet/addModule()}} ### {#add-module-monkey-patch}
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1141,7 +1147,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
     The <dfn method for="WindowSharedStorage">createWorklet(|moduleURL|, |options|)</dfn> method steps are:
 
     1. Let |sharedStorageWorklet| be a new {{SharedStorageWorklet}}.
-    1. Set |sharedStorageWorklet|'s [=cross-origin worklet allowed=] to true.
+    1. If |options| [=map/contains=] |dataOrigin|, set |sharedStorageWorklet|'s [=data origin=] to |options|(|dataOrigin|).
```
xyaoinum, line 1150: `|options|(|dataOrigin|)` --> `|options|[|dataOrigin|]`

spec.bs
```diff
@@ -270,21 +272,21 @@ When {{Worklet/addModule()}} is called for a worklet, it will run [=check if add
...
-    1. Let |allowedInOpaqueOriginContext| be |isCrossOriginWorklet|.
-    1. If the result of running [=determine whether shared storage is allowed by context=] given the [=current settings object=], |workletOrigin|, and |allowedInOpaqueOriginContext| is false, return "DisallowedDueToNonPreferenceError".
-    1. Set |worklet|'s [=SharedStorageWorklet/is cross-origin worklet=] to |isCrossOriginWorklet|.
-    1. If |worklet|'s [=cross-origin worklet allowed=] is false, and if |worklet|'s [=SharedStorageWorklet/is cross-origin worklet=] is true, return "DisallowedDueToNonPreferenceError".
-    1. If the result of running [=check if user preference setting allows access to shared storage=] given the [=current settings object=] and |workletOrigin| is false, return "DisallowedDueToPreferenceError".
+    1. Let |workletDataOrigin| be the [=current settings object=]'s [=environment settings object/origin=].
+    1. If |worklet|'s [=data origin=] is `"script-origin"`, let |workletDataOrigin| be |moduleURLRecord|'s [=url/origin=].
```
xyaoinum, line 283: Two comments:
- `let |workletDataOrigin| be` --> `set |workletDataOrigin| to`.
- `[=data origin=]` --> `[=SharedStorageWorklet/data origin=]`.

spec.bs
```diff
@@ -270,21 +272,21 @@ When {{Worklet/addModule()}} is called for a worklet, it will run [=check if add
...
     - For [[#ss-fetch-algo]], |environment| is the request's [=request/window=], and |origin| is the request's [=request/current URL=]'s [=url/origin=].
-    - For [[#ss-fetch-algo]], for {{WindowSharedStorage/createWorklet()}} called with a cross-origin worklet script, and for {{SharedStorageWorklet/selectURL()}} and {{SharedStorageWorklet/run()}} that operate on a cross-origin worklet created from {{WindowSharedStorage/createWorklet()}}, |allowedInOpaqueOriginContext| is true. For other methods, |allowedInOpaqueOriginContext| is false.
+    - For [[#ss-fetch-algo]], for {{WindowSharedStorage/createWorklet()}} called with a cross-origin worklet script using the <var ignore=''>dataOrigin</var> option with value `"script-origin"` (which would result in a worklet where [=has cross-origin data origin=] is true), and for {{SharedStorageWorklet/selectURL()}} and {{SharedStorageWorklet/run()}} that operate on a worklet where [=has cross-origin data origin=] is true, |allowedInOpaqueOriginContext| is true. For other methods, |allowedInOpaqueOriginContext| is false.
```
xyaoinum, line 275: `[=has cross-origin data origin=]` --> `[=SharedStorageWorklet/has cross-origin data origin=]` for both references

xyaoinum (commented):
Thanks! Added a few nit comments. However, I think the following two sections also need to be updated:
- https://wicg.github.io/shared-storage/#set-up-a-worklet-environment-settings-object-monkey-patch. Instead of always returning `|moduleURL|'s [=url/origin=]`, it may also return the context origin depending on the options.
- The network response's handling: https://wicg.github.io/shared-storage/#http-fetch-monkey-patch. e.g. it doesn't need to do the response header checking if the option is the default "context-origin".

spec.bs

spec.bs

spec.bs

pythagoraskitty:
> Thanks! Added a few nit comments. However, I think the following two sections also need to be updated:
> 
> * https://wicg.github.io/shared-storage/#set-up-a-worklet-environment-settings-object-monkey-patch. Instead of always returning `|moduleURL|'s [=url/origin=]`, it may also return the context origin depending on the options.
> * The network response's handling: https://wicg.github.io/shared-storage/#http-fetch-monkey-patch. e.g. it doesn't need to do the response header checking if the option is the default "context-origin".

Will update these too, thanks!

spec.bs
```diff
@@ -443,7 +445,9 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
         <b>The [=environment settings object/origin=]</b>
           1. Let |workletGlobalScope| be the [=global object=] of <var ignore=''>realmExecutionContext</var>'s Realm component.
           1. If |workletGlobalScope| is not {{SharedStorageWorkletGlobalScope}}, return |origin|.
-          1. Let |pendingAddedModules| be a [=list/clone=] of |worklet|'s [=added modules list=].
+          1. [=Assert=] that |worklet| is a {{SharedStorageWorklet}}.
+          1. If |worklet|'s [=SharedStorageWorklet/data origin=] is `"context-origin"`, return <var ignore=''>outsideSettings</var>'s [=environment settings object/origin=].
+          1. Else let |pendingAddedModules| be a [=list/clone=] of |worklet|'s [=added modules list=].
```
xyaoinum, line 450: `|pendingAddedModules|` should be defined outside the if/else scope

xyaoinum (commented):
Looks good % I'm not entirely convinced that relying on destination "sharedstorageworkletcontextorigin" is ideal (also as noted in my [Gerrit comment](https://chromium-review.googlesource.com/c/chromium/src/+/5648386/comments/fe0fca80_134551fa)). I recommend getting input from a spec owner.

spec.bs

pythagoraskitty:
> Looks good % I'm not entirely convinced that relying on destination "sharedstorageworkletcontextorigin" is ideal (also as noted in my [Gerrit comment](https://chromium-review.googlesource.com/c/chromium/src/+/5648386/comments/fe0fca80_134551fa)). I recommend getting input from a spec owner.

Thanks, will do. I will either need to add the additional `worklet destination type`, or else a new field to `request`, so that I can distinguish requests that we need to check for the "Shared-Storage-Cross-Origin-Worklet-Allowed" header from ones that we don't. As we don't currently have access to the data origin itself from the `request`, we either need to pass that, or some bit denoting whether it's cross-origin or not.

xyaoinum (approved):

wanderview:
> > Looks good % I'm not entirely convinced that relying on destination "sharedstorageworkletcontextorigin" is ideal (also as noted in my [Gerrit comment](https://chromium-review.googlesource.com/c/chromium/src/+/5648386/comments/fe0fca80_134551fa)). I recommend getting input from a spec owner.
> 
> Thanks, will do. I will either need to add the additional `worklet destination type`, or else a new field to `request`, so that I can distinguish requests that we need to check for the "Shared-Storage-Cross-Origin-Worklet-Allowed" header from ones that we don't. As we don't currently have access to the data origin itself from the `request`, we either need to pass that, or some bit denoting whether it's cross-origin or not.

I've reviewed this pull request in regards to the destination question.  I think it would be preferable to monkey-patch an internal property on `request` instead.

My reasoning is that "how do we know when to add the header" is an internal implementation detail.  In contrast, changing the destination is a change that results in different behavior for web sites.  The destination value is exposed to sites a few different ways.  So changing the destination for this purpose would require all sites checking the destination to know have to understand an additional enumerated value.

In addition, using an internal `request` property is consistent with how the fetch spec tracks whether various headers need to be added.

Another option you could consider, though, is sending the worklet data origin explicitly in a request header.  This would maybe benefit the server receiving the request and also give you a comparison you can make to know to add the other header.  (Or do you need the other header if you have this?)  But I can see reasons why maybe not to do that.

wanderview:
Also, just for my curiosity, what is the use case for using the origin of the addWorklet URL as the data origin?  The linked explainer pull request https://github.com/WICG/shared-storage/pull/158 says cross-origin worklet scripts are motivated by CDN usage.  I don't understand a scenario where someone would want data partitioned according to CDN origin.

Is there another use case that isn't documented here?

xyaoinum:
> Also, just for my curiosity, what is the use case for using the origin of the addWorklet URL as the data origin?

The use case for allowing the origin of the script URL as the dataOrigin is to enable third-party scripts to load worklets directly, without requiring an iframe. This is a pre-existing use case (https://github.com/WICG/shared-storage/pull/131) and isn't related to CDN. That is, an ad script could call `createWorklet('https://ad.com/script.js')` in any context. Without this API, they'd need to create an `https://ad.com` iframe first, and inside the iframe, call `addModule('script.js')`.

However, neither of the two approaches supports hosting the script on a CDN, as cross-origin script hasn't been allowed for `addModule()`. This PR removes this limitation, so that CDN script is supported, although it would definitely require running inside the ad iframe.

To align `addModule()` and `createWorlet()`, this PR also changes `createWorklet()`'s default dataOrigin to the calling context's origin. As a result, the network stack needs a separate flag to know _cross-origin & shared-storage & script-origin_ case, whereas before, _cross-origin & shared-storage_ could always imply _script-origin_.

xyaoinum:
(@pythagoraskitty  Re-requested change per Ben's comment. To monkey-patch an internal property on request, you can refer to the [Topics spec](https://patcg-individual-drafts.github.io/topics/#send-browsing-topics-header-boolean-associated-with-request-header))

wanderview (approved):
LGTM with comment addressed.

spec.bs
```diff
@@ -460,28 +467,43 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
 
-  Cross-origin worklets rely on CORS as a baseline permission mechanism to indicate trusted external origins. However, CORS alone is insufficient for worklet creation. Unlike simple resource sharing, worklets allow the creator site to execute JavaScript within the context of the target origin. To ensure security, an additional header, \`<a http-header><code>Shared-Storage-Cross-Origin-Worklet-Allowed</code></a>\`, is required.
+  ### Monkey Patch for [=fetch a single module script=] ### {#fetch-a-single-module-script-monkey-patch}
+  The following step will be added to the [=fetch a single module script=] steps, after setting the request's destination (i.e. "9. Set |request|'s [=request/destination=] to the result..."):
+
+  10. If <var ignore=''>destination</var> is "sharedstorageworklet" and <var ignore=''>settingsObject</var>'s [=environment settings object/origin=] is [=opaque origin|opaque=], then [=map/set=] <var ignore=''>moduleMap</var>[(<var ignore=''>url</var>, <var ignore=''>moduleType</var>)] to null, run <var ignore=''>onComplete</var> given null, and abort these steps.
```
pythagoraskitty, line 489: I added lines 486-489 to address [your comment](https://github.com/WICG/shared-storage/pull/161/files#r1679841312). Let me know if you think this check resolves the issue. Thanks!

spec.bs
```diff
@@ -460,28 +467,43 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
       1. [=Assert=]: |request|'s [=request/origin=] is not "<code>client</code>".
       1. If |request|'s [=request/origin=] and |request|'s [=request/URL=]'s [=url/origin=] are not [=same origin=]:
-          1. Let |headers| be |internalResponse|'s [=response/header list=].
-          1. Let |allowed| be the result of running [=get a structured field value=] algorithm given \`<a http-header><code>Shared-Storage-Cross-Origin-Worklet-Allowed</code></a>\`, "item", and |headers| as input.
-          1. If |allowed| is false, then return a [=network error=].
+          1. Let |dataOriginValue| be the result of [=header list/getting=] `"Sec-Shared-Storage-Data-Origin"` from |request|'s [=request/header list=].
+          1. If |dataOriginValue| is not null, then:
+              1. Let |dataOriginUrl| be the result of running a [=URL parser=] on |dataOriginValue|.
+              1. [=Assert=] that |dataOriginUrl| is not failure.
+              1. If |dataOriginUrl|'s [=url/origin=] and |request|'s [=request/URL=]'s [=url/origin=] are [=same origin=]:
```
xyaoinum, line 501: You only set the header when data-origin is cross-origin to the context origin, then it seems we don't need to compare the origins here. Should we instead assert that `|dataOriginUrl|'s [=url/origin=]` equals `|request|'s [=request/URL=]'s [=url/origin=]`?

spec.bs
xyaoinum: I think this isn't necessary: It seems "dataOrigin is not opaque" is already guaranteed by the check in https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/161.html#determine-whether-shared-storage-is-allowed-by-context, and if addModule fails early, it won't fetch the script.

wanderview: LGTM

wanderview (approved):
LGTM

spec.bs
pythagoraskitty: Revised as suggested. 

Note that the `[=Assert=] that |dataOriginUrl|'s [=url/origin=] and |request|'s [=request/URL=]'s [=url/origin=] are [=same origin=].` will become a conditional for the remaining steps if we eventually allow other data origins besides the context origin and the script origin.

spec.bs
pythagoraskitty: Removed the patch for fetch a single module script, since [determine-whether-shared-storage-is-allowed-by-context](https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/161.html#determine-whether-shared-storage-is-allowed-by-context) performs the needed check. Thanks, @xyaoinum !

PR #167, opened 2024-08-05
Make run() signature consistent for selectURL

ankwok:
Some examples have run(data, urls) whereas the formal spec has run(urls, data) as the signature.

https://wicg.github.io/shared-storage/#get-the-select-url-result-index


pythagoraskitty (approved):

pythagoraskitty:
@ankwok Thank you for catching this!

PR #168, opened 2024-08-07
Use `sharedStorage` instead of `this.sharedStorage` in worklet

pythagoraskitty:
In worklet code,  use `sharedStorage` directly inside of the class methods instead of `this.sharedStorage`, as the latter is incorrect. 

`this` will refer to the operation object class rather than the worklet. `globalThis` would be needed instead to refer to the worklet, but `sharedStorage` can simply be referred to directly here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ class SelectURLOperation {
   hash(experimentName, seed) { … }
 
   async run(urls, data) {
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
@xyaoinum PTAL, thanks!

xyaoinum (approved):

PR #169, opened 2024-08-09
Adjust callback function invocation.

jeremyroman:
Unforunately this isn't purely editorial, because in the process of adjusting the use of "invoke" a number of other things needed to be clarified.

Namely:
* the operation map now holds JS function objects (i.e., anything with [[Call]])
* these function objects are converted to a specific IDL types only when they are about to be invoked
* the select URL algorithm reacts to the promise rather than expecting its value (or rejection reason) to be immediately available

Part of whatwg/webidl#1425.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/shared-storage/pull/169.html" title="Last updated on Aug 9, 2024, 8:17 PM UTC (d838e10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/169/52b4ff5...jeremyroman:d838e10.html" title="Last updated on Aug 9, 2024, 8:17 PM UTC (d838e10)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -71,7 +71,6 @@ spec: webidl; urlPrefix: https://webidl.spec.whatwg.org
         text: async iterator; url: idl-async-iterable
         text: promise; url: idl-promise
         text: promise rejected; url: a-promise-rejected-with
-        text: convert; for: ecmascript-to-idl; url: dfn-convert-ecmascript-to-idl-value
 spec: storage; urlPrefix: https://storage.spec.whatwg.org/
     type: dfn
         text: storage model; url: model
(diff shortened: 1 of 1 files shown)


jeremyroman:
@pythagoraskitty ptal? I think this is what is intended here (modulo some other things that remain a little unclear but aren't directly blocking the invoke change).

spec.bs
```diff
@@ -420,13 +425,11 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
             1. If |operationMap| [=map/contains=] |name|:
-                1.  Let |operation| be |operationMap|[|name|].
-                1. [=Assert=]: |operation|'s [=associated realm=] is [=this=]'s [=relevant realm=].
-                1. If |options| [=map/contains=] |data|:
-                    1. Let |argumentsList| be a new [=/list=].
-                    1. [=list/Append=] |data| to |argumentsList|.
-                    1. [=Invoke=] |operation| with |argumentsList|.
-                1. Otherwise, [=invoke=] |operation| without any arguments list.
+                1. [=Assert=]: |operationMap|[|name|]'s [=associated realm=] is [=this=]'s [=relevant realm=].
+                1. Let |operation| be |operationMap|[|name|], [=converted to an IDL value|converted=] to {{Function}}.
```
xyaoinum, line 429: Quick question (tangential to the CL but related): In the code, we define both [RunFunctionForSharedStorageRunOperation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_global_scope.idl?q=RunFunctionForSharedStorageRunOperation) and [RunFunctionForSharedStorageSelectURLOperation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_global_scope.idl?q=RunFunctionForSharedStorageSelectURLOperation). But in the spec, only `RunFunctionForSharedStorageSelectURLOperation` is defined. This is because for `RunFunctionForSharedStorageRunOperation`, the end behavior seems the same as calling the operation as a `{{Function}}`. I'm wondering if we should strive for consistency here.

xyaoinum (approved):
LGTM % a quick question.

spec.bs
jeremyroman: Since nothing actually waits on the promise returned from the operation (intentionally, according to the note), I suspect the difference is not web-observable. It would probably be good to make the two consistent. I'd rather not block this PR on that, but all else equal, easy correspondence between implementations and spec makes it easier to evolve the two in parallel.

jeremyroman:
When this is ready in your opinion, I think one of you will need to merge it, as I don't have write access to this repository.

PR #178, opened 2024-09-03
[Explainer] Mention `resolveToConfig` in list of supported options

pythagoraskitty:
`resolveToConfig` was already described as a supported option for `selectURL()` in the explainer, but we omitted to list in the bullet list of supported options. We correct this previous omission. We also slightly clarify the list to make it apparent which options are supported for `run()` and which for `selectURL()`.


pythagoraskitty:
@xyaoinum PTAL, thanks!

xyaoinum (approved):

PR #176, opened 2024-09-03
[Explainer] Support Saved Queries in selectURL

pythagoraskitty:
`selectURL()` currently has two per-page-load budgets that restrict that number of calls made to `selectURL()` on each page-load.  We propose allowing queries to be saved and reused on a per-page basis, where the per-page-load budgets are charged the first time a saved query is run but not for subsequent runs of the saved query during the same page-load. This will be accomplished with a `savedQuery` parameter in the options for `selectURL()` that will name the query.

PR #188, opened 2024-09-10
[spec] Support saved queries for selectURL

pythagoraskitty:
We update the spec to support saved queries, as a followup to #176.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/188.html" title="Last updated on Sep 20, 2024, 5:14 PM UTC (fa8183f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/188/0ebd7a5...fa8183f.html" title="Last updated on Sep 20, 2024, 5:14 PM UTC (fa8183f)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -872,7 +872,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
   <dl dfn-for="bit debit">
     : <dfn>bits</dfn>
-    :: a double
+    :: a {{double}}
 
     : <dfn>timestamp</dfn>
     :: a {{DOMHighResTimeStamp}} (from the [=Unix Epoch=])
@@ -958,7 +958,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
   <dl dfn-for="shared storage page budget">
     : <dfn>overall budget</dfn>
-    :: a double
+    :: a {{double}}
 
     : <dfn>site budget map</dfn>
     :: a [=map=] of [=site=] to {{double}}
--- a/spec.bs
+++ b/spec.bs
@@ -955,7 +955,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
     :: a [=map=] of [=site=] to {{double}}
 
     : <dfn>saved query map</dfn>
-    :: a [=map=] of [=tuples=] ([=url/origin=], [=/URL=], [=string=], [=string=]) to {{unsigned long}}
+    :: a [=map=] of [=tuples=] ([=url/origin=] <dfn for="saved query map" export>data origin</dfn>, [=/URL=] <dfn for="saved query map" export>worklet script URL</dfn>, [=string=] <dfn for="saved query map" export>operation name</dfn>, [=string=] <dfn for="saved query map" export>query name</dfn>) to {{unsigned long}}
   </dl>
 
 
--- a/spec.bs
+++ b/spec.bs
@@ -334,9 +334,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
     1. If |window|'s [=associated document=] is not [=fully active=], return a [=promise rejected=] with a {{TypeError}}.
     1. Let |savedQueryName| be |options|["`savedQuery`"].
     1. If |savedQueryName| is a [=string=] that is not the empty string, then:
-        1. Let |savedIndex| be the result of running [=get the index for a saved query=] on |navigable|, |workletDataOrigin|, and |savedQueryName|.
-
-            Note: We intentionally ignore |operationName| when running [=get the index for a saved query=], in order to allow the names of saved queries to be independent from the names of operations.
+        1. Let |savedIndex| be the result of running [=get the index for a saved query=] on |navigable|, |workletDataOrigin|, |operationName|, and |savedQueryName|.
         1. If |savedIndex| is not null, then:
             1. If |savedIndex| is greater than |urlList|'s [=list/size=], return a [=promise rejected=] with a {{TypeError}}.
 
@@ -366,9 +364,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
                         Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
 
                     1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
-                    1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |navigable|, |workletDataOrigin|, |savedQueryName|, and |index|.
-
-                        Note: We intentionally ignore |operationName| when running [=store the index for a saved query=], in order to allow the names of saved queries to be independent from the names of operations.
+                    1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |navigable|, |workletDataOrigin|, |operationName|, |savedQueryName|, and |index|.
                     1. Run |privateAggregationCompletionTask|.
 
                 :   If it was rejected:
@@ -967,7 +963,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
     :: a [=map=] of [=site=] to {{double}}
 
     : <dfn>saved query map</dfn>
-    :: a [=map=] of [=url/origin=] to [=map=] of [=string=] to {{unsigned long}}
+    :: a [=map=] of [=url/origin=] to [=map=] of [=string=] to [=map=] of [=string=] to {{unsigned long}}
   </dl>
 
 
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
@xyaoinum Could you please take a first pass? I won't land until after #176 lands and after I have a spec mentor review also. Thanks!

spec.bs
```diff
@@ -1250,6 +1285,7 @@ On the other hand, methods for getting data from the [=shared storage database=]
     boolean resolveToConfig = false;
     boolean keepAlive = false;
     SharedStoragePrivateAggregationConfig privateAggregationConfig;
+    DOMString savedQuery = "";
```
xyaoinum, line 1288: I think usually we omit the `= ""` for string default values.

xyaoinum (approved):
LGTM % a nit

spec.bs

pythagoraskitty:
@wanderview Please take a look, thanks!

spec.bs
```diff
@@ -324,13 +325,22 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+    To <dfn>get the select-url result index</dfn>, given {{SharedStorageWorklet}} |worklet|, {{DOMString}} |operationName|, [=/list=] of {{SharedStorageUrlWithMetadata}}s |urlList|, an [=/origin=] |workletDataOrigin|, a [=/navigable=] |navigable|, {{SharedStorageRunOperationMethodOptions}} |options|, a [=pre-specified report parameters=] or null |preSpecifiedParams| and an [=aggregation coordinator=] or null |aggregationCoordinator|:
 
     1. Let |promise| be a new [=promise=].
     1. Let |window| be |worklet|'s [=relevant settings object=].
     1. [=Assert=]: |window| is a {{Window}}.
     1. If |window|'s [=Window/browsing context=] is null, then return a [=promise rejected=] with a {{TypeError}}.
     1. If |window|'s [=associated document=] is not [=fully active=], return a [=promise rejected=] with a {{TypeError}}.
+    1. Let |savedQueryName| be |options|["`savedQuery`"].
+    1. If |savedQueryName| is nonempty, then:
```
wanderview, line 336: Is "nonempty" a spec term?  Is there somewhere else that is using this?  Maybe you want to say `is a string that is not the empty string`?

spec.bs
```diff
@@ -324,13 +325,22 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+        1. Let |site| be the result of running [=obtain a site=] on |workletDataOrigin|.
+        1. Let |savedIndex| be the result of running [=get the index for a saved query=] on |navigable|, |site|, and |savedQueryName|.
```
wanderview, line 338: It appears that you could use the same name for two completely different queries and this would return the value of the first query.  Is this the desired behavior?

For example, maybe we should return an error if the parameters passed to selectUrl in the second call differ from the first call?  Or should the second call override the parameters from the first selectUrl by creating a new saved query?

This is kind of stemming from the awkardness of adding a parameter to an existing function call.  It seems like what you might want is something more like how SQL has separate "prepare statement" and "execute statement" calls.  You could consider a `prepareSelectUrl()` that takes all params and assigns a name.  Then a separate `selectUrl(name)` call that operates only on the name without any other params.

If we want to keep a single function call that is overloaded I think we should just be intentional if the site calls it twice with different arguments, but the same name.

wanderview (changes requested):

spec.bs
pythagoraskitty: > It appears that you could use the same name for two completely different queries and this would return the value of the first query. Is this the desired behavior?

The short answer is: probably yes.  

The long answer is that I did forget to take into account the operation name, and I may not want to ignore that. I am thinking about that some more offline.  

But we deliberately want the array of URLs to be able to be different, and note that we do handle the error that happens if we end up with an out of range index because of this.

We would ignore any passed `data` for `options` in the case that we are able to retrieve a saved query result. The other `options` parameters are still handled.

So there shouldn't be a need for a separate `prepareSelectUrl()` statement.

wanderview: In that case I recommend returning an error if the operation name doesn't match the previously used operation name for the same saved query.  Its probably a developer mistake.  Allowing the same saved query name for different operations sounds like a footgun.

spec.bs

spec.bs
pythagoraskitty: After further contemplation and discussion with the team, we've decided two things.

1. Saved queries will be stored on a per-origin (instead of per-site) basis. I have updated this spec PR accordingly.
2. We do indeed want names of saved queries to be independent of operation names, and for the storing and lookup of saved queries to ignore the operation names. So two calls to `selectURL()` that have different operation names could still access the same saved query result if they use the same `savedQuery` parameter. While this could lead to some developers making mistakes, it also allows for greater flexibility and simplicity. We would need to be sure to document this properly, of course, including in the current explainer PR (where the effect of differing operation names is currently not mentioned).

jkarlin: I don't have particularly strong feelings here. It seems unlikely to be a footgun to me, but at the same time I can't think of a good reason to use the same saved-query on two different operations. So, let's say that we go ahead with your proposal here as it seems reasonable.

So now the proposal is that a saved-query key is `<origin, operation_name, saved_query_name>`. 

For similar reasons, what would you think about also including the worklet script url as part of the key?

e.g., `<origin, operation_name, worklet_script_url, saved_query_name>`

Like operations, presumably different scripts could have different behavior, even if the operation name is the same. This gets a little funky in that urls aren't actually const resources, but still.

pythagoraskitty: OK, I have updated the spec so that queries are keyed by `<origin, operation_name, saved_query_name>`.

I think we should just not reuse the query if there is different operation name, rather than throwing, to simplify things.

wanderview: Just to be clear, keying on `<origin, operation_name, saved_query_name>` does your proposal above, right?  It ignores mismatched operations because you end up with different slots in your map.  To do what I was proposing I think the key needs to be `<origin, saved_query_name>` and then in the map value the `operation_name` is saved so you can compare it on lookup.

I'm not sure I understand enough about how different worklets can mix on a page to have an opinion on how to treat conflicting scripts URLs.  If we assume folks are going to hash URLs for different versions of the same worklet, though, it does seem like a flow we should be concerned with.

pythagoraskitty: > Just to be clear, keying on <origin, operation_name, saved_query_name> does your proposal above, right? It ignores mismatched operations because you end up with different slots in your map. To do what I was proposing I think the key needs to be <origin, saved_query_name> and then in the map value the operation_name is saved so you can compare it on lookup.

Yes. The key would need to be `<origin, saved_query_name` to do what you propose. But I think it would be best to key on `<origin, operation_name, saved_query_name>` instead because it seems to make more sense for us to simply narrow the scope for the saved queries further from the beginning, then to add additional errors. We wouldn't throw if there is a mismatch, but we wouldn't re-use a query in the case of a mismatch either.

spec.bs
```diff
@@ -974,6 +987,29 @@ navigables</a> section, add the following:
...
+    1. If |topLevelTraversable|'s [=page budget=]'s [=saved query map=][|origin|][|operationName|] does not [=map/contain=] |savedQueryName|, then return null.
+    1. Return |topLevelTraversable|'s [=page budget=]'s [=saved query map=][|origin|][|operationName|][|savedQueryName|].
+  </div>
+
+  <div algorithm>
+    To <dfn>store the index for a saved query</dfn>, given [=/navigable=] |navigable|, [=url/origin=] |origin|, [=string=] |operationName|, [=string=] |savedQueryName|, and {{unsigned long}} |index|:
+
+    1. Let |topLevelTraversable| be the result of running [=get the top-level traversable=] for |navigable|.
+    1. [=Assert=] that |topLevelTraversable|'s [=page budget=] is not null.
+    1. If |topLevelTraversable|'s [=page budget=]'s [=saved query map=] does not [=map/contain=] |origin|, then [=map/set=] |topLevelTraversable|'s [=page budget=]'s [=saved query map=][|origin|] to an [=map/empty=] [=map=].
```
wanderview, line 1008: Rather than a map containing a map that contains a further map, you can instead use a tuple as the map key.  For example:

```
[=map/Set=] |topLevelTraversable|'s [=page budget=]'s [=saved query map=][(|origin|, |operationName|, |savedQueryName|)] to |index|.
```

wanderview (approved):
LGTM with one spec language improvement fixed.  This changes the setting and getting of the saved query using tuples as the key.

spec.bs
wanderview: An example of tuples as keys here: https://html.spec.whatwg.org/multipage/webappapis.html#module-map

wanderview: LGTM

pythagoraskitty:
After further discussion with the team, we decided to also key on the script's URL. So now I have updated the spec so the the key is `<origin, script URL, operation name, query name>`.

Would you mind please taking one more look? @wanderview @xyaoinum 

Thanks!

spec.bs
```diff
@@ -937,10 +949,13 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
     : <dfn>overall budget</dfn>
-    :: a double
+    :: a {{double}}
 
     : <dfn>site budget map</dfn>
-    :: a [=map=] of [=sites=] to doubles
+    :: a [=map=] of [=site=] to {{double}}
+
+    : <dfn>saved query map</dfn>
+    :: a [=map=] of [=tuples=] ([=url/origin=], [=/URL=], [=string=], [=string=]) to {{unsigned long}}
```
wanderview, line 958: I recommend phrasing this like `a [=string=] <dfn for="saved query map">operation name</dfn>, a [=string=] <dfn for="saved query map">saved query name< /dfn>`, etc.  So folks can tell the difference between the strings, etc.  I would do it for the URL and origin as well for completeness.

wanderview:
> After further discussion with the team, we decided to also key on the script's URL. So now I have updated the spec so the the key is `<origin, script URL, operation name, query name>`.
> 
> Would you mind please taking one more look? @wanderview @xyaoinum
> 
> Thanks!

I added one more comment, but otherwise LGTM.

xyaoinum:
Still LGTM

spec.bs

PR #197, opened 2024-09-19
[explainer] Web Locks integration

xyaoinum:
Today, concurrent execution of shared storage worklets in scenarios like cross-site reach measurement can result in duplicate reporting, due to the potential race conditions within the "get() and set()" logic. To address this issue, we propose integrating the Web Locks API into Shared Storage.

Specifically:
- Inside the worklet: introduce `navigator.locks.request` similar to the existing Web Locks API.
- For modifier methods (set, append, delete, clear): introduce a `withLockOnResource` option to allow handling the operation with the lock held.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,21 @@ Shared Storage is a key/value store that is partitioned by calling origin (but n
 
 See the [draft specification](https://wicg.github.io/shared-storage/).
 
-### Demonstration 
+## APIs built on top of Shared Storage
+
+This document only describes the core shared storage framework and infrastructure to store cross-site data privately and to read that data from within a secure worklet environment. APIs that use shared storage's data to produce some output are linked to below.
+
+### Private Aggregation
+
+The [Private Aggregation](https://github.com/patcg-individual-drafts/private-aggregation-api) API allows for aggregated histograms to be sent based on data read from shared storage. The histograms are differentially private.
+
+
+### Select URL
+
+The [selectURL](https://github.com/WICG/shared-storage/blob/main/select-url.md) API allows for content selection based on cross-site data. It takes 8 possible URLs as input and sends them to a worklet which selects from a small list of URLs. The chosen URL is stored in a fenced frame config as an opaque form that can only be read by a [fenced frame](https://github.com/WICG/fenced-frame); the embedder does not learn this information. 
+
+
+## Demonstration 
 
 You can [try out](https://shared-storage-demo.web.app/) Shared Storage along with some APIs built for it using Chrome 104+.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -267,6 +277,45 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
 *  The response header will only be honored if the corresponding request included the request header: `Sec-Shared-Storage-Writable: ?1`.
 *  See example usage below.
 
+### Locking for Modifier Methods
+
+All modifier methods (`set`, `append`, `delete`, `clear`), whether invoked from JavaScript or from response headers, accept a `withLockOnResource: <resource>` option. This option instructs the method to acquire a lock on the designated resource before executing.
+
+The locks requested this way are partitioned by the shared storage data origin, and are independent of any locks obtained via `navigator.locks.request` in a Window or Worker context. Note that they share the same scope with the locks obtained via `navigator.locks.request` in the SharedStorageWorklet context.
+
+Unlike `navigator.locks.request` that offers various configuration options, the locks requested this way always use the default settings:
```
pythagoraskitty, line 286: s/Unlike `navigator.locks.request` that offers various configuration options,/Unlike `navigator.locks.request`, which offers various configuration options,/

README.md
```diff
@@ -267,6 +277,45 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
+Developers should be mindful of this limitation. To get around, consider one of these options:
+1.  Move the modifier methods inside the worklet. This allows a single lock to protect them.
+2.  Place the data under the same key and modify it within a single method, which is inherently atomic.
+
+### Recommendations for lock usage
+
+If the worklet performs only a single read or write, the lock is unnecessary (both inside and outside the worklet) because each read or write is inherently atomic. Example: [A/B experiment](#simple-example-consistent-ab-experiments-across-sites).
+
+If the worklet performs both read and write on the same key, the lock is likely necessary. Examples: [cross-site reach measurement](#cross-site-reach-measurement), [creative selection by frequency](#creative-selection-by-frequency).
+
```
pythagoraskitty, line 318: I like the examples you have below of locks inside the worklet. Could we perhaps have a contextual example of why we would want to use a lock outside the worklet? 

(I know you have the example showing that ordering may be unexpected, but perhaps an example that motivates why one would want to use the option `withLockOnResource` in the first place?)

xyaoinum: I came up with a naive example. I didn't come up with a more meaningful one: This would involve the worklet doing multiple read and writes, but:
- If the worklet only interacts with a single key (like the "reach" example), it's unclear in what scenarios the key also needs modification outside the worklet.
- If the worklet operates on multiple keys, it likely encounters the limitations outlined in [Caveat #2](https://github.com/WICG/shared-storage/blob/yao-web-locks/README.md#caveat-2-no-atomicity-for-multiple-writes), which makes it unsuitable as a good example.

README.md
xyaoinum: Done.

README.md
pythagoraskitty: Thank you for adding the example!

pythagoraskitty (approved):
LGTM, but since I'm a drive-by, you probably want @jkarlin to also take a look :)

PR #198, opened 2024-09-20
[Explainer] Support custom `dataOrigin` in createWorklet

pythagoraskitty:
We have developers currently testing the Shared Storage API who have indicated they want to be able to run their worklet on a different origin from the origin that wrote the data. In particular, one of these two actions may be served by, e.g., a CDN, and the other by another domain.  

Third-party cookies allow sharing data at the site level. The current architecture of shared storage, however, only allows the invoking context's origin for `addModule()`/`createWorklet()` to be used as the shared storage data partition origin, or in the case of `createWorklet()`, the worklet script's origin may be manually selected as the data partition origin using the option `dataOrigin: "script-origin"`.

We suggest making it possible for a worklet script to use shared storage data from an arbitrary origin other than the worklet script's origin or the invoking context's origin. The data origin would have to opt-in for data use to occur, as data processing could modify the data origin's data, send reports, and would affect its shared storage privacy budget. 

While the worklet script will be able to process the data origin's data within a shared storage worklet, the current restrictions will remain in place on output of shared storage data.


README.md
```diff
@@ -171,11 +171,36 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
+        *   To use a custom origin as the partition origin for accessing shared storage, pass the `dataOrigin` option with the serialized partition origin (i.e. the [URL serialization](https://url.spec.whatwg.org/#url-serializing) of a URL with the same scheme, host, and port as the partition origin, but whose other components are empty) as its value in the `options` dictionary.
+        *   Supported values for the `dataOrigin` option, if used, are the keywords "script-origin" and "context-origin", as well as any valid serialized HTTPS origin, e.g. "https://custom-data-origin.example". 
+            *   "script-origin" designates the worklet script origin as the data partition origin.
+            *   "context-origin" (default) designates the invoking context origin as the data partition origin.
+            *   If a valid serialized HTTPS URL is passed as the value for `dataOrigin`, then the origin of the URL parsed from the `dataOrigin` string will be used as the worklet's data partition origin, as long as the URL's origin has opted-in via a [/.well-known/](#well-known) to specify that the worklet script's origin and context origin are trusted.
```
xyaoinum, line 178: Suggest to keep it brief here, e.g. "a serialized HTTPS origin designates itself as the data partition origin." And then, right next to this paragraph, add the paragraph "When a valid serialized HTTPS URL is passed as the value for dataOrigin ..", which explains how the data origin is treated. This removes some of the redundancy.

README.md
```diff
@@ -171,11 +171,36 @@ The shared storage worklet invocation methods (`addModule`, `run`, and `selectUR
...
     *   The object that the returned Promise resolves to has the same type with the implicitly constructed `window.sharedStorage.worklet`. However, for a worklet created via `window.sharedStorage.createWorklet(url, options)`, only `selectURL()` and `run()` are available, whereas calling `addModule()` will throw an error. This is to prevent leaking shared storage data via `addModule()`, similar to the reason why `addModule()` can only be invoked once on the implicitly constructed `window.sharedStorage.worklet`.
     *   Redirects are not allowed.
-    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin and when `dataOrigin` is "script-origin", a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
+    *   When the module script's URL's origin is cross-origin with the worklet's creator window's origin and when `dataOrigin` is "script-origin" (or is a valid serialized HTTPS URL that is same-origin to the worklet's script's origin), a `Shared-Storage-Cross-Origin-Worklet-Allowed: ?1` response header is required.
```
xyaoinum, line 181: Do we still require the ./well-known mechanism in this case? I think either one of the following is reasonable:  1) we skip the ./well-known check completely, and just rely on the script's response header. 2) if either mechanism indicates "allowed", then we'll allow it.

In either case, let's clearly document the mechanism.

README.md

README.md
pythagoraskitty: I was thinking we do (1), skip the /.well-known/ check completely, and just rely on the script's response header.

I will make that explicit.

README.md
```diff
@@ -165,11 +165,36 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+        *   Supported values for the `dataOrigin` option, if used, are the keywords "script-origin" and "context-origin", as well as any valid serialized HTTPS origin, e.g. "https://custom-data-origin.example". 
+            *   "script-origin" designates the worklet script origin as the data partition origin.
+            *   "context-origin" (default) designates the invoking context origin as the data partition origin.
+            *   A serialized HTTPS origin designates itself as the data partition origin.
+    *   When a valid serialized HTTPS URL is passed as the value for `dataOrigin` and the parsed URL's origin is cross-origin to both the invoking context's origin and the worklet script's origin, the parsed URL's origin must host a JSON file at the <a name="well-known">/.well-known/</a> path "/.well-known/shared-storage/trusted-origins" with an array of dictionaries, each with keys `scriptOrigin` and `contextOrigin`. The values for these keys should be either a string or an array of strings.
+        *   A string value should be either a serialized origin or "*", where "*" matches all origins.
```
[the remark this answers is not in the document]
pythagoraskitty: Fixed.

PR #199, opened 2024-09-27
[explainer] Web Locks integration

xyaoinum:
Today, concurrent execution of shared storage worklets in scenarios like cross-site reach measurement can result in duplicate reporting, due to the potential race conditions within the "get() and set()" logic. To address this issue, we propose integrating the Web Locks API into Shared Storage.

Specifically,
- Introduce `withLock: <resource>` option to all modifier methods.
- Introduce `navigator.locks.request` to the worklet environment.
- Introduce a batch modify method: `sharedStorage.batchUpdate(methods, options)`. This method, with the `withLock` option, allows multiple modifier methods to be executed atomically, enabling use cases where a website needs to maintain consistency while updating data organized across multiple keys.


README.md
```diff
@@ -138,12 +138,19 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
     *   `key` and `value` are both strings.
     *   Options include:
         *   `ignoreIfPresent` (defaults to false): if true, a `key`’s entry is not updated if the `key` already exists. The embedder is not notified which occurred.
-*   `window.sharedStorage.append(key, value)`
+        * `withLockOnResource`: acquire a lock on the designated resource before executing. See [Locking for modifier methods](#locking-for-modifier-methods) for details.
```
jkarlin, line 141: I'm partial to `withLock`. If we need a third word then I'd prefer, `withLockRequest` as that is closer to the lockmanager api.

README.md
```diff
@@ -206,6 +213,9 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
             *   The approximate size of the contents of this interest group, in bytes.
     *   The [AuctionAdInterestGroup](https://wicg.github.io/turtledove/#dictdef-auctionadinterestgroup)'s [lifetimeMs](https://wicg.github.io/turtledove/#dom-auctionadinterestgroup-lifetimems) field will remain unset. It's no longer applicable at query time and is replaced with attributes `timeSinceGroupJoinedMs` and `lifetimeRemainingMs`.
     *   This API provides the Protected Audience buyer with a better picture of what's happening with their users, allowing for Private Aggregation reports.
+*   `navigator.locks.request(resource, callback)` and `navigator.locks.request(resource, options, callback)`
```
jkarlin, line 216: Does `navigator.locks` also return a LockManager as it does in a document?

README.md
```diff
@@ -206,6 +213,9 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+    *   Acquires a lock on `resource` and invokes `callback` with the lock held. See the [request](https://w3c.github.io/web-locks/#dom-lockmanager-request) method in Web Locks API for details.
+    *   Lock Scope: shared storage locks are partitioned by the shared storage data origin, and are independent of any locks obtained via `navigator.locks.request` in a Window or Worker context.
```
jkarlin, line 218: Please mention that this is for privacy reasons. It should likely also get a sentence or two in the privacy section.

README.md
```diff
@@ -251,6 +262,77 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+sharedStorage.set('key0', 'value2');
+```
+
+Developers should be mindful of this potential ordering issue.
+
+#### Caveat 2: No atomicity for multiple writes
+
+The `withLockOnResource` option provides locking for individual methods but doesn't guarantee atomicity across multiple methods.
+
+Developers should be mindful of this limitation. To get around, consider one of these options:
```
jkarlin, line 326: Suggest s/To get around,/To work around it,/

README.md
```diff
@@ -251,6 +262,77 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+1.  Move the modifier methods inside the worklet. This allows a single lock to protect them.
```
jkarlin, line 327: Needs a bit more description, e.g., that you need to send the data via the data parameter to the worklet.

README.md
```diff
@@ -251,6 +262,77 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+All modifier methods (`set`, `append`, `delete`, `clear`), whether invoked from JavaScript or from response headers, accept a `withLockOnResource: <resource>` option. This option instructs the method to acquire a lock on the designated resource before executing.
+
+The locks requested this way are partitioned by the shared storage data origin, and are independent of any locks obtained via `navigator.locks.request` in a Window or Worker context. Note that they share the same scope with the locks obtained via `navigator.locks.request` in the SharedStorageWorklet context.
+
+Unlike `navigator.locks.request`, which offers various configuration options, the locks requested this way always use the default settings:
+*  `mode: "exclusive"`: The lock is never shared with other locks.
+*  `steal: false`: The lock will not preempt other locks.
+*  `ifAvailable: false`: If the lock is currently held by others, keep waiting and don't skip.
+
+#### Example: Naive Integrity Check
```
jkarlin, line 276: I see the race here, but it's pretty unclear what the purpose is.

Perhaps instead, write to three variables, a, b, and c from the document and within the worklet read all three and add a comment that if some document (which could be in another tab!) is writing while the worklet is reading, then they may not be consistent.

README.md
```diff
@@ -251,6 +262,77 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+
+#### Caveat 1: Unexpected ordering
+
+Modifier methods may block due to the lock, so may not execute in the order they appear in the code.
+
+```js
+// Resolve immediately. Internally, this may block to wait for the lock to be granted.
+sharedStorage.set('key0', 'value1', { withLockOnResource: 'resource0' });
+
+// Resolve immediately. Internally, this will also execute immediately.
```
jkarlin, line 316: What do you mean by 'also' here?

select-url.md
```diff
@@ -158,14 +158,16 @@ class CreativeSelectionByFrequencyOperation {
...
 
-    let count = await sharedStorage.get(data.campaignId);
-    count = count ? parseInt(count) : 0;
-
-    // If under cap, return the desired ad.
-    if (count < 3) {
-      index = 1;
-      sharedStorage.set(data.campaignId, (count + 1).toString());
-    }
+    await navigator.locks.request("creation-selection-by-frequency-lock", async (lock) => {
```
jkarlin, line 161: Suggest adding a comment above as to why you're creating the lock.

README.md
xyaoinum: Done.

README.md
xyaoinum: Yes. Clarified in the text.

README.md
xyaoinum: Done.

README.md
xyaoinum: As I mentioned in [a previous thread](https://github.com/WICG/shared-storage/pull/197#discussion_r1767828720), I'm finding it difficult to write a compelling example for this. Two concerns:

1. Regarding your three-variable example, what specific use case does it illustrate? (Reach? A/B?) What do the variables represent? We can certainly leave it vague.
2. If consistency is crucial, I suspect the use case will encounter [Caveat 2: No atomicity for multiple writes](https://github.com/WICG/shared-storage/blob/yao-web-locks/README.md#caveat-2-no-atomicity-for-multiple-writes). This could make the example less effective.

Ultimately, I believe this locking mechanism will likely be ineffective without a "multi-write" operation. PLMK if you still prefer updating the example. Thanks!

README.md
xyaoinum: 'also' refers to the 'immediately' part. I admit it's ambiguous. Removed.

select-url.md
xyaoinum: Done.

xyaoinum:
@jkarlin I've addressed most of your comments % still struggling with a good example for `withLock`. PTAL. Thanks!

jkarlin (approved):

README.md
```diff
@@ -214,18 +231,22 @@ The shared storage worklet invocation methods (`addModule`, `createWorklet`, and
...
+        *   `set(<key>, <value>, {withLock: <resource>})` &larr;&rarr; `set;key=<key>;value=<value>;with_lock=<resource>`
         *   `set(<key>, <value>)` &larr;&rarr; `set;key=<key>;value=<value>`
         *   `append(<key>, <value>)` &larr;&rarr; `append;key=<key>;value=<value>`
         *   `delete(<key>)` &larr;&rarr; `delete;key=<key>`
         *   `clear()` &larr;&rarr; `clear`
+        *   `batchUpdate(<methods>, {withLock: <resource>})` &larr;&rarr; `batchUpdate;methods=<serialized methods>;with_lock=<resource>`
+            *   `<serialized methods>` is a [String Item](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings) containing the [List Structured Header](https://www.rfc-editor.org/rfc/rfc8941.html#name-lists) serialization of batched methods in the same list structure. However, nested `batchUpdate()` operations within a `batchUpdate()` are not allowed.
+            *   Example `<serialized methods>`: `"set;key=\"123\";value=\"456\";ignore_if_present, append;key=abc;value=def"`. This corresponds to a `[new SharedStorageSetMethod("123", "456", {ignoreIfPresent: true}), new SharedStorageAppendMethod("abc", "def")]` argument in JavaScript.
```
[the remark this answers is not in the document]
xyaoinum: Done.

PR #202, opened 2024-10-07
[spec] Update saved query to be async

pythagoraskitty:
We fix our spec changes so that a query can be queued to reused if it is not the initial query but it is received before the initial query is resolved.

Previously in the spec, only fully resolved queries were being treated as saved. This was an oversight.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/202.html" title="Last updated on Oct 28, 2024, 10:32 PM UTC (a532f49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/202/8121c4e...a532f49.html" title="Last updated on Oct 28, 2024, 10:32 PM UTC (a532f49)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -369,7 +369,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
                         Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
 
-                    1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
+                    1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with a [=tuple=] (|index|, true).
                     1. Run |privateAggregationCompletionTask|.
 
                 :   If it was rejected:
@@ -434,12 +434,14 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
     1. If |options|["`resolveToConfig`"] is true, [=resolve=] |resultPromise| with |pendingConfig|.
     1. Otherwise, [=resolve=] |resultPromise| with |urn|.
     1. Let |indexPromise| be the result of running [=get the select-url result index=], given [=this=], |name|, |urlList|, |workletDataOrigin|, |navigable|, |options|, |preSpecifiedParams| and |aggregationCoordinator|.
-    1. [=Upon fulfillment=] of |indexPromise| with |resultIndex|, perform the following steps:
+    1. [=Upon fulfillment=] of |indexPromise| with (|resultIndex|, |shouldChargeTopLevelBudgets|), perform the following steps:
         1. Let |site| be the result of running [=obtain a site=] with |document|'s [=Document/origin=].
         1. Let |remainingBudget| be the result of running [=determine remaining navigation budget=] with |environment| and |site|.
         1. Let |pendingBits| be the logarithm base 2 of |urlList|'s [=list/size=].
-        1. let |pageBudgetResult| be the result of running [=charge shared storage top-level traversable budgets=] with |navigable|, |site|, and |pendingBits|.
-        1. If |pageBudgetResult| is false, or if |pendingBits| is greather than |remainingBudget|, set |resultIndex| to the [=default selectURL index=].
+        1. If |shouldChargeTopLevelBudgets| is true:
+            1. Let |pageBudgetResult| be the result of running [=charge shared storage top-level traversable budgets=] with |navigable|, |site|, and |pendingBits|.
+            1. If |pageBudgetResult| is false, set |resultIndex| to the [=default selectURL index=].
+        1. If |pendingBits| is greather than |remainingBudget|, set |resultIndex| to the [=default selectURL index=].
         1. Let |finalConfig| be a new [=fenced frame config=].
         1. Set |finalConfig|'s [=fenced frame config/mapped url=] to |urlList|[|resultIndex|].
         1. Set |finalConfig|'s <span class=todo>a "pending shared storage budget debit" field</span> to |pendingBits|.
@@ -1032,7 +1034,7 @@ navigables</a> section, add the following:
     1. Let |processIndexTask| be an algorithm to perform the following steps, given an {{unsigned long}} |index|:
         1. If |index| is greater than |urlList|'s [=list/size=], then [=queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}.
                 Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
-        1.  Otherwise, [=queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
+        1.  Otherwise, [=queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with the [=tuple=] (|index|, false).
     1. Return |processIndexTask|.
   </div>
 
--- a/spec.bs
+++ b/spec.bs
@@ -375,9 +375,10 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
 
                         Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
 
-                    1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |index|.
-                    1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
-                    1. Run |privateAggregationCompletionTask|.
+                        Otherwise:
+                        1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |index|.
+                        1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
+                        1. Run |privateAggregationCompletionTask|.
 
                 :   If it was rejected:
                 ::  1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and the [=default selectURL index=].
--- a/spec.bs
+++ b/spec.bs
@@ -369,7 +369,8 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
                     1. Run |privateAggregationCompletionTask|.
 
                 :   If it was rejected:
-                ::  1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}.
+                ::  1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and 0.
+                    1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}.
 
                         Note: This indicates that either |operationCtor|'s run() method encounters an error (where |operationCtor| is the parameter in {{SharedStorageWorkletGlobalScope/register()}}), or the result |index| is a non-integer value, which violates the selectURL() protocol, and we don't know which url should be selected.
                     1. Run |privateAggregationCompletionTask|.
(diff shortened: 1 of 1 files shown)


pythagoraskitty:
@xyaoinum and @wanderview Please take a look, thanks!

spec.bs
```diff
@@ -360,12 +360,12 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+                ::  1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |index|.
+                    1. If |index| is greater than |urlList|'s [=list/size=], then [=queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}, and abort these steps.
 
                         Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
 
                     1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=resolve=] |promise| with |index|.
-                    1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |index|.
                     1. Run |privateAggregationCompletionTask|.
 
                 :   If it was rejected:
```
xyaoinum, line 371: What should happen to the waiting callbacks if the first selectURL() is rejected (before validating index against urlList’s [size](https://infra.spec.whatwg.org/#list-size))? I don't see they are being handled. IMO, we can/should also cache this response, and return the default index 0.

For queryData’s [index](https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/202.html#saved-query-data-index), you might want to distinguish between "unresolved" and "rejected". Or, perhaps you can reuse "0" for the "rejected" status.

spec.bs
```diff
@@ -982,20 +991,40 @@ navigables</a> section, add the following:
   #### Saved queries #### {#saved-queries}
 
   <div algorithm>
```
xyaoinum, line 993: Suggest adding a brief note below this algorithm, describing what the return value means. (-2, -1, >=0)

xyaoinum (commented):
Thanks! Have one question about the behavior for when the first selectURL() is rejected.

spec.bs
pythagoraskitty: Let's reuse "0" for the rejected status.

Added below in lines 372-373.

spec.bs
pythagoraskitty: Done.

xyaoinum:
Two more questions on budget charging:
1. How did you avoid double-charging the per-page budget when returning a saved state? It looks like they will all end up in step "33. [Upon fulfillment](https://webidl.spec.whatwg.org/#upon-fulfillment) of indexPromise with resultIndex ...", which will always charge the budget. Did I miss anything?
2. It appears that the budget is not being charged in "34. [Upon rejection](https://webidl.spec.whatwg.org/#upon-rejection) of indexPromise, ...", which seems incorrect and a pre-existing bug? Suggest either filing a separate issue to track, or fixing it directly here. The fix seems simple: just mirror the logic used when the promise is fulfilled % needing to use default index 0.

pythagoraskitty:
> Two more questions on budget charging:
> 
> 1. How did you avoid double-charging the per-page budget when returning a saved state? It looks like they will all end up in step "33. [Upon fulfillment](https://webidl.spec.whatwg.org/#upon-fulfillment) of indexPromise with resultIndex ...", which will always charge the budget. Did I miss anything?

You're correct; there were some bugs in my logic that I'v now fixed. First, i forgot to return |promise| and abort in the case that we already had a saved index available (I implied that this is what I wanted done but didn't mention it explicitly.) Second, I was still charging the budgets, as you noted. Now I will use a tuple with a boolean to keep track of whether or not to charge the budgets.

> 2. It appears that the budget is not being charged in "34. [Upon rejection](https://webidl.spec.whatwg.org/#upon-rejection) of indexPromise, ...", which seems incorrect and a pre-existing bug? Suggest either filing a separate issue to track, or fixing it directly here. The fix seems simple: just mirror the logic used when the promise is fulfilled % needing to use default index 0.

This issue I don't quite follow. Why do we need to charge budget in the rejection case? Won't the charge be 0 bits? What did I miss?

xyaoinum:
LGTM % the rejection case question

> This issue I don't quite follow. Why do we need to charge budget in the rejection case? Won't the charge be 0 bits? What did I miss?

If we don't charge the budget for reject, then if selectURL() happens to return 0, it could instead trigger an error to achieve the desired outcome without incurring budget costs. This issue was previously addressed in the code: https://crrev.com/c/3846383 (which seems to occur before we had the spec)

pythagoraskitty:
> LGTM % the rejection case question
> 
> > This issue I don't quite follow. Why do we need to charge budget in the rejection case? Won't the charge be 0 bits? What did I miss?
> 
> If we don't charge the budget for reject, then if selectURL() happens to return 0, it could instead trigger an error to achieve the desired outcome without incurring budget costs. This issue was previously addressed in the code: https://crrev.com/c/3846383 (which seems to occur before we had the spec)

I had forgotten that we had done that. Thanks for reminding me. I will fix the rejection case now in the spec.

pythagoraskitty:
@xyaoinum Please take another look, as I think I fixed the rejection bug. Thanks!

xyaoinum (approved):

spec.bs
```diff
@@ -324,27 +327,31 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
   ## Run Operation Methods on {{SharedStorageWorklet}} ## {#run-op-shared-storage-worklet}
 
   <div algorithm>
-    To <dfn>get the select-url result index</dfn>, given {{SharedStorageWorklet}} |worklet|, {{DOMString}} |operationName|, [=/list=] of {{SharedStorageUrlWithMetadata}}s |urlList|, an [=/origin=] |workletDataOrigin|, a [=/navigable=] |navigable|, {{SharedStorageRunOperationMethodOptions}} |options|, a [=pre-specified report parameters=] or null |preSpecifiedParams| and an [=aggregation coordinator=] or null |aggregationCoordinator|:
+    To <dfn>get the select-url result index</dfn>, given {{SharedStorageWorklet}} |worklet|, {{DOMString}} |operationName|, [=/list=] of [=strings=] |urlList|, an [=/origin=] |workletDataOrigin|, a [=/navigable=] |navigable|, {{SharedStorageRunOperationMethodOptions}} |options|, a [=pre-specified report parameters=] or null |preSpecifiedParams| and an [=aggregation coordinator=] or null |aggregationCoordinator|:
```
wanderview, line 330: I think we need to document the boolean in the tuple returned as right now its very hard to tell what the `true` and `false` values mean.  Maybe phrase this first paragraph as "... run the following steps.  This algorithm will return a tuples consisting of a promise that results to the select URL and a boolean indicating if the budget should be charged".  Or something to that effect.

Using a string with enum values instead of a boolean would also be more readable.

spec.bs
```diff
@@ -982,20 +1004,44 @@ navigables</a> section, add the following:
...
+        1. Set |queryData|'s [=saved query data/index=] value to -1.
+        1. Return -2.
+    1. Let |savedIndex| be |topLevelTraversable|'s [=page budget=]'s [=saved query map=][(|origin|, |moduleURLRecord|, |operationName|, |savedQueryName|)]'s  [=saved query data/index=].
+    1. If |savedIndex| is -1:
+        1. [=queue/Enqueue=] |callbackTask| to |queryData|'s [=saved query data/callbacks=].
+        1. Return -1.
+    1. Return |savedIndex|.
   </div>
 
+      Note: The [=get the index for a saved query=] algorithm returns -2 to indicate that [=obtain a worklet agent|a worklet agent should be obtained=] and the index value is pending the result of the worklet agent's operation.
```
wanderview, line 1022: Rather than `-1` and `-2` as magic numbers, I wonder if you should just use fixed strings here.  It would be more readable and since there is no strict return type I think its allowed.

spec.bs
```diff
@@ -324,27 +327,31 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
     1. [=Assert=]: |moduleMapKeyTuples| has [=map/size=] 1.
     1. Let |moduleURLRecord| be |moduleMapKeyTuples|[0][0].
     1. Let |savedQueryName| be |options|["`savedQuery`"].
     1. If |savedQueryName| is a [=string=] that is not the empty string, then:
-        1. Let |savedIndex| be the result of running [=get the index for a saved query=] on |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, and |savedQueryName|.
-        1. If |savedIndex| is not null, then:
-            1. If |savedIndex| is greater than |urlList|'s [=list/size=], return a [=promise rejected=] with a {{TypeError}}.
+        1. Let |callbackTask| be the result of running [=obtain a callback to process the saved index result=], given |window|, |urlList|, and |promise|.
+        1. Let |savedIndex| be the result of running [=get the index for a saved query=] on |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |callbackTask|.
+        1. If |savedIndex| is -1, then return the [=tuple=] (|promise|, false) and abort these steps.
```
wanderview, line 346: Is there a reason you are using `return` and `abort these steps` here?  If its an async algorithm then I don't think return necessarily makes sense, but if its a sync algorithm then abort probably doesn't make sense.  I don't see where it says to run the steps in parallel, though, so it seems to be a sync algorithm, right?

spec.bs
```diff
@@ -428,25 +457,9 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
-        1. [=Finalize a pending config=] on |fencedFrameConfigMapping| with |urn| and |finalConfig|.
-        1. Let |resultURLWithMetadata| be |urls|[|resultIndex|].
-        1. If |resultURLWithMetadata| has field "`reportingMetadata`", run [=register reporting metadata=] with |resultURLWithMetadata|["`reportingMetadata`"].
-        1. If |options|["`keepAlive`"] is false, run [=terminate a worklet global scope=] with [=this=].
-    1. [=Upon rejection=] of |indexPromise|, perform the following steps:
-        1. Let |finalConfig| be a new [=fenced frame config=].
-        1. Set |finalConfig|'s [=fenced frame config/mapped url=] to |urlList|[[=default selectURL index=]].
-        1. [=Finalize a pending config=] on |fencedFrameConfigMapping| with |urn| and |finalConfig|.
-        1. If |options|["`keepAlive`"] is false, run [=terminate a worklet global scope=] with [=this=].
+    1. Let (|indexPromise|, |shouldChargeTopLevelBudgets|) be the result of running [=get the select-url result index=], given [=this=], |name|, |urlList|, |workletDataOrigin|, |navigable|, |options|, |preSpecifiedParams| and |aggregationCoordinator|.
```
wanderview, line 460: It seems `|shouldChargeTopLevelBudget|` is not used until after the promise is resolved.  It might be easier to work with this algorithm if the promise resolves to a tuple {URL, boolean} rather than having {promise, boolean}.  What do you think?

wanderview (changes requested):

spec.bs
pythagoraskitty: I would like the boolean to be available even in the case where the promise is rejected. That's why it seems like I need to use {promise, boolean}, instead of resolving the entire promise to a tuple {unsigned long, boolean}.

spec.bs
pythagoraskitty: This returns a tuple {promise, boolean}, so I believe it's async, but it could be fulfilled synchronously. 

I don't quite follow what you are saying, though. We need to return the {promise, boolean} and abort the rest of the steps. What are you suggesting that we do instead?

spec.bs
pythagoraskitty: Annotated the algorithm to describe the boolean returned.

spec.bs

spec.bs
pythagoraskitty: OK, I believe I understand. 

So in this case, we are in the sync part of the algorithm and so early returns will cut the steps short. I have removed "abort...".

spec.bs
```diff
@@ -324,27 +327,32 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+        1. If |savedIndex| is "pending callback", then return the [=tuple=] (|promise|, false).
+
+            Note: |callbackTask| is now stored to be run when a previously obtained worklet agent completes its operation to select the index for this query. When the steps of |callbackTask| are run, |promise| will be resolved.
+        1. If |savedIndex| is an {{unsigned long}}, then:
+            1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to run the steps of |callbackTask|, given |savedIndex|.
+
+                Note: Running the steps of |callbackTask| will resolve |promise|.
+            1. Return the [=tuple=] (|promise|, false).
+        1. [=Assert=] that |savedIndex| is "obtain worklet agent".
+    1. Return the [=tuple=] (|promise|, true), and immediately [=obtain a worklet agent=] given |window| and run the rest of these steps in that agent:
```
wanderview, line 355: I haven't seen this "return foo and then do something async" style before.  I think it would be preferred to structure it like this:

```
1. [=Obtain a worklet agent=] given |window| and run the following steps in that agent:
  1. // fill in async steps
1. Return the [=tuple=] (|promise|, true).
```

The async steps will effectively continue to run after the return here.

wanderview: Although I think we should `[=queue a task=]` on the worklet agent's event loop instead of using the vague "run the steps in that agent".

See:

https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task
https://html.spec.whatwg.org/#concept-agent-event-loop

spec.bs
```diff
@@ -360,20 +369,45 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+                ::  1. If |index| is greater than |urlList|'s [=list/size=], then:
+                        1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and the [=default selectURL index=].
+                        1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}, and abort these steps.
 
                         Note: The result index is beyond the input urls' size. This violates the selectURL() protocol, and we don't know which url should be selected.
 
+                    1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and |index|.
```
xyaoinum, line 378: Should this step and the next one be put in an "else" clause following the previous "if"?

wanderview (approved):

spec.bs
pythagoraskitty: Done, except I used "Otherwise" instead of "Else", as seems to be what the HTML standard uses.

PR #205, opened 2024-10-31
Handle Shared-Storage-Write as a single batchUpdate(); Respect withLo…

xyaoinum:
…ck for inner methods within batchUpdate()

Two revisions to our initial proposal:

1. For `Shared-Storage-Write`, our initial proposal included individual methods like set(), append(), delete(), and clear(), alongside a batchUpdate() method requiring a serialized string for its `methods` parameter. To make things simpler, we revise the proposal to treat everything as a single batchUpdate(). It's backward compatible.

3. Our initial proposal ignored the `withLock` option for individual methods within a batchUpdate() operation. This approach lacked flexibility and could lead to unintended behavior, as developers might assume the lock should still be acquired. We've revised the proposal to respect the `withLock` option for each individual method within the batch.


xyaoinum:
@pythagoraskitty / @jkarlin : PTAL. Thanks!

pythagoraskitty:
Should there be any updates to the [examples section for response headers](https://github.com/WICG/shared-storage?tab=readme-ov-file#writing-to-shared-storage-via-response-headers)?

Otherwise I think it is looking good % my question above. I will take another pass if you make more changes. :)

xyaoinum:
> Should there be any updates to the [examples section for response headers](https://github.com/WICG/shared-storage?tab=readme-ov-file#writing-to-shared-storage-via-response-headers)?

Updated to using `batchUpdate()` (although the original example should also be valid, as a `batchUpdate()` without locks should be equivalent to executing individual methods in sequence). I also included two examples within the [From response headers](https://github.com/WICG/shared-storage/tree/yao-update-web-locks-header?tab=readme-ov-file#from-response-headers) section.

README.md
```diff
@@ -466,16 +471,20 @@ On the client side, to initiate the request:
...
-sharedStorage.clear();
-sharedStorage.set("hello", "world", {ignoreIfPresent: true});
-sharedStorage.append("good", "bye");
-sharedStorage.delete("hello");
-sharedStorage.set("all", "done");
+
+sharedStorage.batchUpdate([
+  new SharedStorageClearMethod(),
+  new SharedStorageSetMethod("hello", "world", {ignoreIfPresent: true}),
+  new SharedStorageAppendMethod("good", "bye"),
```
jkarlin, line 483: Suggest adding a lock to one of these ops to show that it can be done.

jkarlin (approved):
lgtm w/ comment

README.md
xyaoinum: Done.

pythagoraskitty (approved):

PR #207, opened 2024-11-14
[spec] Remove interestGroups()'s permissions policy check

xyaoinum:
No need to check for permission "run-ad-auction" or "join-ad-interest-group":
- The ability to join or to run auctions are different from the ability to view the data.
- Relying on the existing "private-aggregation" permission check at the reporting time should be sufficient.


xyaoinum:
@pythagoraskitty: PTAL. Thanks!

pythagoraskitty (approved):

PR #216, opened 2025-01-09
[spec] Enable per-context contribution limits for Private Aggregation

dmcardle:
The goal is to enable Shared Storage embedders to override the default number of contributions per Private Aggregation report.

To that end, this change adds the `maxContributions` field to the web-visible Private Aggregation config dictionary and plumbs its value into Private Aggregation's "pre-specified report parameters".

Context:
* Explainer: patcg-individual-drafts/private-aggregation-api#146
* Spec change: patcg-individual-drafts/private-aggregation-api#164

***
<a href="https://pr-preview.s3.amazonaws.com/dmcardle/shared-storage/pull/216.html" title="Last updated on Jan 16, 2025, 4:28 PM UTC (e98524b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/216/6df142d...dmcardle:e98524b.html" title="Last updated on Jan 16, 2025, 4:28 PM UTC (e98524b)">Diff</a>

PR #219, opened 2025-01-16
[spec] Make bikeshed check references to Private Aggregation API

dmcardle:
Currently, all references to the private-aggregation-api spec are declared in the "anchors" block. According to the bikeshed docs, this is how you link to dfns in specs that aren't part in the autolinking database[^1]. As a result, bikeshed hasn't actually been checking our references, nor has it always been generating the correct links[^2].

Private Aggregation spec is already in the database[^3], so I think we can promote these to actual references.

This commit makes the necessary adjustments to the "link-defaults" block. With this change, invalid references to the Private Aggregation spec are now compilation errors.

[^1]: https://speced.github.io/bikeshed/#custom-dfns
[^2]: For instance, clicking "context ID" in step 10 of "obtain the pre-specified report parameters" links to a non-existent anchor on the Private Aggregation spec.
[^3]: https://github.com/tobie/specref/blob/3569d1ecfa298461f3de0586ed2d13ffb000fa29/refs/browser-specs.json#L1179


dmcardle:
@alexmturner and @xyaoinum, PTAL!

(Ignore the first commit, this PR is based on #216.)

spec.bs
```diff
@@ -8,6 +8,7 @@ Repository: WICG/shared-storage
 URL: https://github.com/WICG/shared-storage
 Editor: Camillia Smith Barnes, Google https://google.com, cammie@chromium.org
 Markup Shorthands: markdown yes
+Text Macro: PRIVATE-AGGREGATION-API [[private-aggregation-api|Private Aggregation API]]
```
alexmturner, line 11: Wondering if it might be clearer to just put the full link in rather than defining a macro

spec.bs
```diff
@@ -579,11 +572,20 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
         1. If |filteringIdMaxBytes| is not the [=default filtering ID max bytes=] or
             |contextId| is not null, return a new {{DOMException}} with name
             "`DataError`".
+    1. Let |maxContributions| be null.
```
alexmturner, line 575: We may want to define this a PR into the other PR so it doesn't include both changes

alexmturner (commented):
LGTM w nits

spec.bs
dmcardle: Good point, forgot to call that out. I've added `DO_NOT_MERGE` to the PR title for the time being.

spec.bs
dmcardle: Fair enough! It's just that Bikeshed already knows the URL, so it feels unnecessarily verbose to repeat it.

I've seen other specs do something like this:

```md
Something something <cite>Private Aggregation API</cite>. [[!PRIVATE-AGGREGATION-API]
```

For instance, see how the [Fetch](https://fetch.spec.whatwg.org/#websocket-protocol) spec refers to WebSockets.

The `<cite>` tag doesn't seem to get autolinked, but it's rendered in italics. Bikeshed does, however, linkify the part in square brackets. The Bikeshed docs say that `[[FOO]]` is for normative references and `[[!FOO]]` is for informative references, so I suppose the context matters there.

dmcardle:
Rebased, this is now disentangled from the #216, modulo the fact that it imports the `max contributions` dfn (but never references it).

spec.bs
dmcardle: I should call out that I've now made some editorial decisions that enabled me to avoid explicitly linking to Private Aggregation's spec. I think it's actually a little clearer this way, too, because we autolink to the relevant dfns instead of dropping the reader on the front door of the Private Aggregation spec.

spec.bs
```diff
@@ -124,36 +147,6 @@ spec: permissions-policy; urlPrefix: https://www.w3.org/TR/permissions-policy/
 spec: attestation; urlPrefix: https://github.com/privacysandbox/attestation
     type: dfn
         text: enrolled
-spec: private-aggregation-api; urlPrefix: https://patcg-individual-drafts.github.io/private-aggregation-api/
-    type: dfn
-        text: Private Aggregation; url:
```
dmcardle, line 129: This little tidbit is what enabled us to link to the Private Aggregation spec with `[=Private Aggregation=]`. It's kind of a hack though -- it works by pretending there's a dfn with that name, then overriding the value of the URL hash as the empty string. Technically, it kind of got the job done, but I suppose it risks colliding with something that Private Aggregation actually wanted to export.

spec.bs
```diff
@@ -267,7 +260,7 @@ This document introduces a new storage API that is intentionally not partitioned
 
 The {{SharedStorageWorklet}} Interface {#worklet}
 =================================================
-The {{SharedStorageWorklet}} object allows developers to supply [=module scripts=] to process [=Shared Storage=] data and then output the result through one or more of the output gates. Currently there are two output gates, the [=Private Aggregation=] output gate and the {{SharedStorageWorklet/selectURL()|URL-selection}} output gate.
+The {{SharedStorageWorklet}} object allows developers to supply [=module scripts=] to process [=Shared Storage=] data and then output the result through one or more of the output gates. Currently there are two output gates, the <code>{{privateAggregation}}.{{PrivateAggregation/contributeToHistogram()}}</code> output gate and the {{SharedStorageWorklet/selectURL()|URL-selection}} output gate.
```
alexmturner, line 263: I understand the desire to link to a definition, but we're introducing contributeToHistogramOnEvent(), so I think it might be better to just link to the spec overall. We could just do that with an <a> tag, though.

spec.bs
alexmturner: Definitely a hack, but I don't think there's a real risk of collision here.

dmcardle: Specifically, I was thinking that mixing and matching link-defaults and anchors might cause occurrences of `[=Private Aggregation=]` to incorrectly autolink to the permissions policy string, declared as [`<dfn>private-aggregation</dfn>`](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/3433a3e8ff358e3b93a3b57b31ce63204bf8afb4/spec.bs#L494), either now or in a future version of bikeshed.

spec.bs
dmcardle: Works for me :)

dmcardle: I went with the normative reference syntax (`[[!FOO]]`).

spec.bs
```diff
@@ -2400,4 +2393,4 @@ Privacy Considerations {#privacy}
 
   In particular, an embedder can select a [=/URL=] from a short list of [=/URL=]s based on data in their shared storage and then display the result in a [=fenced frame=]. The embedder will not be able to know which [=/URL=] was chosen except through specific mechanisms that will be better-mitigated in the longer term. Currently, a few bits of entropy can leak each time that the user clicks on the [=fenced frame=] to initiate a [=top-level traversable=] [=navigate|navigation=] and/or the [=fenced frame=] calls the {{reportEvent()}} API.
 
-  An embedder is also able to send aggregatable reports via the [=Private Aggregation=] API, which adds noise in order to achieve differential privacy, uses a time delay to send reports, imposes limits on the number of reports sent, and processes the reports into aggregate data so that individual privacy is protected.
+  An embedder is also able to send aggregatable reports via {{privateAggregation}}, which adds noise in order to achieve differential privacy, uses a time delay to send reports, imposes limits on the number of reports sent, and processes the reports into aggregate data so that individual privacy is protected.
```
alexmturner, line 2396: minor: Might prefer a reference to the API generally here, but up to you

alexmturner (approved):

spec.bs
dmcardle: Done!

spec.bs
alexmturner: Looks like this was undone in the latest force-push

dmcardle: Good catch! Fixed.

PR #221, opened 2025-01-21
Fix HTML errors noticed by bikeshed-5.0.

jyasskin:


pythagoraskitty:
Hello, looks like this came for my review while I was on leave. Will take a look in the next couple days.

pythagoraskitty:
Ah, looks like this has been fixed already by now in https://github.com/WICG/shared-storage/commit/142b73557b647b6c41cbf6607a20af777ca7ffac.

Thank you @dmcardle !

PR #224, opened 2025-01-28
[spec] Fix syntax errors detected by bikeshed 5.0.0

dmcardle:
This addresses the following errors by adding missing `<div algorithm>` opening tags:

    LINE 1519: Saw an end tag </div>, but there's no open element corresponding to it.
    LINE 1540: Saw an end tag </div>, but there's no open element corresponding to it.
    LINE 1559: Saw an end tag </div>, but there's no open element corresponding to it.

Next, the following error appeared, which revealed we've been using the wrong linking syntax for `this` on line 1561 and in many other places.

    LINE 1561: The var 'this' (in algorithm 'SharedStorageClearMethod()' for SharedStorageClearMethod) is only used once.


dmcardle:
@jkarlin PTAL!

PR #227, opened 2025-02-12
[spec] During addModule(), switch to allow `navigator` but disallow `navigator.locks`

xyaoinum:
Currently, `navigator.locks` is the only available attribute on `navigator` (i.e., a SharedStorageWorkletNavigator), and we want to prevent Web Locks functionality during addModule().

This change avoids unintentionally breaking existing code that might legitimately read the navigator object (e.g., expects no Exception thrown) during module loading for purposes unrelated to Web Locks.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/pull/227.html" title="Last updated on Feb 12, 2025, 11:56 PM UTC (1cb8bcb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/shared-storage/227/d0d2cb4...1cb8bcb.html" title="Last updated on Feb 12, 2025, 11:56 PM UTC (1cb8bcb)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -103,6 +103,7 @@ spec: web-locks; urlPrefix: https://w3c.github.io/web-locks/
         text: lock manager; url: lock-manager
         text: obtain a lock manager; url: obtain-a-lock-manager
         text: request a lock; url: request-a-lock
+        text: locks getter steps; url: dom-navigatorlocks-locks
 spec: ecma; urlPrefix: https://tc39.es/ecma262/
     type: dfn
         text: call; url: sec-call
@@ -925,8 +926,7 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
   <div algorithm>
     The <dfn attribute for=SharedStorageWorkletGlobalScope>navigator</dfn> [=getter steps=] are:
 
-    1. If [=this=]'s [=addModule success=] is true, return [=this=]'s [=SharedStorageWorkletGlobalScope/navigator instance=].
-    1. Otherwise, throw a {{TypeError}}.
+    1. Return [=this=]'s [=SharedStorageWorkletGlobalScope/navigator instance=].
   </div>
 
   <div algorithm="privateAggregation getter">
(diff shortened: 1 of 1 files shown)


xyaoinum:
@wanderview PTAL. Thanks! (Note that the Web Lock owner approved the behavior change ([code](https://chromium-review.googlesource.com/c/chromium/src/+/6256882))).

wanderview (changes requested):
The changes look reasonable for what they do, but I have a questions/comment.

This "filter out things we don't want" approach tends to create a fragile system because new items can be added to navigator later that you have not considered yet.  Therefore a new storage-like thing could be added and create a security vulnerability.

As an alternative, have you considered listing exactly what attributes on navigator you believe are safe instead?

xyaoinum:
> This "filter out things we don't want" approach tends to create a fragile system because new items can be added to navigator later that you have not considered yet. Therefore a new storage-like thing could be added and create a security vulnerability.
> 
> As an alternative, have you considered listing exactly what attributes on navigator you believe are safe instead?

Thanks for the review @wanderview 

The [ `SharedStorageWorkletNavigator`](https://wicg.github.io/shared-storage/#sharedstorageworkletnavigator) is a newly introduced interface specifically for shared storage worklets. Currently, it's defined to expose only the `locks` attribute. Any future additions would require explicit approval and specification changes by the shared storage team. So I don't see a big risk of unintended exposure.

Regarding _"listing exactly what attributes on navigator you believe are safe instead"_ -- I'm unsure of a normative way to do this within the specification. I imagine we still need to do _some_ checks within the getter methods? And in that case, if we miss a check in any getter, we still introduce the same vulnerability.

wanderview:
Hmm, I see now I misunderstood the goal.  Sorry for my confusion.

Have you considered what this would do to feature detection code that might be running?  What if its trying to feature detect the existence of navigator.lock during the initial evaluation of the script.  Would this break existing scripts in the wild?

It feels like it would be better to make lock manager methods reject immediately if a lock is created or accessed in addModule instead.

WDYT?

xyaoinum:
> Have you considered what this would do to feature detection code that might be running?  What if its trying to feature detect the existence of navigator.lock during the initial evaluation of the script.  Would this break existing scripts in the wild?
> 
> 
> 
> It feels like it would be better to make lock manager methods reject immediately if a lock is created or accessed in addModule instead.
> 
> 
> 
> WDYT?

@wanderview You're right that feature detection could break if errors aren't properly handled or if incorrect assumptions are made about behavior. That's the main motivation behind this change -- moving the throwing logic down a level. However, throwing at the individual method level would make it harder to audit new usage and increase maintenance costs (e.g., the getter methods reside in the shard storage specification, while individual methods are spread across their own specifications).

wanderview:
Ok, if you have considered the tradeoffs then I'm willing to approve.

wanderview (approved):

PR #229, opened 2025-03-07
Spec: Private Aggregation error reporting

alexmturner:
Adds support for the new Private Aggregation error reporting feature to Shared Storage. See the related PAA spec change:
https://github.com/patcg-individual-drafts/private-aggregation-api/pull/172. Also see the explainer:
https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/error_reporting.md

Slightly reorganizes the PAA integration with this spec for readability.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -128,34 +128,6 @@ spec: attestation; urlPrefix: https://github.com/privacysandbox/attestation
 spec: private-aggregation-api; urlPrefix: https://patcg-individual-drafts.github.io/private-aggregation-api/
     type: dfn
         text: Private Aggregation; url:
-        text: get the privateAggregation
-        text: determine if an origin is an aggregation coordinator
-        text: pre-specified report parameters
-        for: pre-specified report parameters
-            text: context ID
-            text: filtering ID max bytes
-            text: max contributions
-        text: batching scope
-        text: debug scope
-        text: process contributions for a batching scope
-        text: set the aggregation coordinator for a batching scope
-        text: determine if a report should be sent deterministically
-        text: mark a debug scope complete
-        text: set the pre-specified report parameters for a batching scope
-        text: aggregation coordinator
-        text: default filtering id max bytes
-        text: valid filtering id max bytes range
-        text: context id
-        text: scoping details
-        for: scoping details
-            text: get batching scope steps
-            text: get debug scope steps
-        text: private-aggregation
-        for: PrivateAggregation
-            text: allowed to use
-            text: scoping details; url: #privateaggregation-scoping-details
-    type: interface
-        text: PrivateAggregation
 spec: protected-audience; urlPrefix: https://wicg.github.io/turtledove/
     type: dfn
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -394,21 +366,27 @@ Moreover, each {{SharedStorageWorklet}}'s [=global scopes|list of global scopes=
...
+                        1. Run |privateAggregationCompletionTask| given [=Private Aggregation completion task trigger/normal completion=].
 
                 :   If it was rejected:
                 ::  1. If |savedQueryName| is a [=string=] that is not the empty string, then run [=store the index for a saved query=] with |window|, |navigable|, |workletDataOrigin|, |moduleURLRecord|, |operationName|, |savedQueryName|, and the [=default selectURL index=].
                     1. [=Queue a global task=] on the [=DOM manipulation task source=], given |window|, to [=reject=] |promise| with a {{TypeError}}.
 
                         Note: This indicates that either |operationCtor|'s run() method encounters an error (where |operationCtor| is the parameter in {{SharedStorageWorkletGlobalScope/register()}}), or the result |index| is a non-integer value, which violates the selectURL() protocol, and we don't know which url should be selected.
-                    1. Run |privateAggregationCompletionTask|.
+                    1. If the promise was rejected as the operation was completed abruptly due to an uncaught exception:
```
alexmturner, line 385: Wasn't sure how to evaluate this properly (in spec language) unless we decide that returning a non-integer should count as an uncaught exception for aggregate error reporting

alexmturner (commented):
@xyaoinum, could you PTAL? Had a question about how best to write this (see comment). Thanks!

Also note that the spec build is failing due to some missing exports, but I have a different PR to fix that in the PAA repo that I'll land first.

spec.bs
xyaoinum: A non-integer will often be implicitly converted to an integer without error, see test case [SelectURL_NonNumericStringConvertedTo0](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_unittest.cc?q=SelectURL_NonNumericStringConvertedTo0). AFAICT, the only possible integer conversion failure is to return an object that defines a custom toString() method that throws, e.g. [SelectURL_ReturnValueToUint32Error](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_unittest.cc?q=SelectURL_ReturnValueToUint32Error). I think it's fine to simply assume 'uncaught exception' here.

xyaoinum (approved):
LGTM

spec.bs
alexmturner: Ah interesting -- I've updated this PR to reflect that / remove the unnecessary if clause. I'll try to add a test case for this situation in the implementation too. Thanks!

alexmturner:
Thanks for the review! Fixed the build and updated the 'uncaught exception' test.

PR #233, opened 2025-04-03
[explainer] Fix: Disallow batch inner locks, don't ignore

xyaoinum:
The explainer incorrectly stated that batch inner locks would be ignored. They are actually disallowed and will cause an error. This commit corrects the explainer.


pythagoraskitty (approved):

pythagoraskitty:
Thank you for fixing this!

